PR #446, opened 2025-04-22
chore: increase auth test coverage

ryanbonial:
### Description

Enhanced the `LoginError` component to handle specific authentication error scenarios:
- Added automatic retry for 401 Unauthorized errors
- Added custom error messages for 404 errors related to invalid session IDs
- Added custom error messages for other 404 errors related to invalid login links

### What to review

- The new test cases for different error scenarios in `LoginError.test.tsx`
- The automatic retry logic for 401 errors
- The specific error message handling for different types of 404 errors
- The updated vitest config exclusions

### Testing

Added comprehensive test coverage for the new error handling scenarios:
- Test for automatic retry on 401 Unauthorized errors
- Test for session ID error messages on specific 404 errors
- Test for login link error messages on other 404 errors

### Fun gif
cover

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  "packages/core": "0.0.2",
-  "packages/react": "0.0.2"
+  "packages/core": "1.0.0",
+  "packages/react": "1.0.0"
 }
--- a/apps/kitchensink-react/package.json
+++ b/apps/kitchensink-react/package.json
@@ -20,7 +20,7 @@
     "@sanity/icons": "^3.7.0",
     "@sanity/sdk-react": "workspace:*",
     "@sanity/ui": "^2.15.13",
-    "groq": "3.86.2-experimental.0",
+    "groq": "3.88.1-typegen-experimental.0",
     "inter-ui": "^4.1.0",
     "json-edit-react": "^1.26.2",
     "react": "^18.3.1",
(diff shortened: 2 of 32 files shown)


ryanbonial:
@ricokahler this is ready for review now

packages/react/src/components/auth/LoginError.test.tsx
```diff
@@ -29,6 +44,100 @@ describe('LoginError', () => {
...
+        method: 'GET',
+        url: 'https://api.sanity.io/some/endpoint',
+        headers: {
+          'content-type': 'application/json',
+        },
+        body: {
+          error: 'Not Found',
+          message: 'Some other 404 error',
+        },
+      }),
```
ricokahler, line 127: it seems somewhat off that this error is not being supplied as the `LoginError`'s `error` prop. I think `react-error-boundary` works by passing the error it catches during render via this prop. i'm a fan of throwing errors in our hooks (instead of propagating them via hook returns) and allowing the error boundary to pick them up instead of us figuring out how it should propgate. this would allow other APIs to also throw a `ClientError`s and this error component would still behave as expected. i believe the current implementation will only show if there is a client error that's set in the auth store where this pattern could work if, for instance, a `useQuery` call threw a ClientError indirectly (as part of just utilizing the sanity client).

no action requested since this PR is only for tests but my preference would be to:
- refactor the selectors in `getAuthState` to throw the error instead of just return it, this would make the error propagate to react
- refactor the `authErrorMessage` in `LoginError` to derive the error message directly from the error passed through props via `'react-error-boundary'`. i don't think it needs to use react state and can be implemented in an effectless/stateless manner
- refactor this test to utilize an `<ErrorBoundary FallbackComponent={LoginError}>` instead of just rendering it

ricokahler (approved):
first time taking a look at the refresh logic with the web locks API. it's some pretty gnarly stuff! i left a comment for ideas on how to potentially improve error handling/propagation but not high priority (or related to the goal of this PR lol)

PR #475, opened 2025-04-30
feat: add new token fetching via comlink

ryanbonial:
### Description

This PR adds Comlink token refresh functionality to enable comlink authentication between the SDK and Dashboard in the case where the app no longer has a valid token and needs a new tokem. It introduces a new `ComlinkTokenRefreshProvider` component and `useComlinkTokenRefresh` hook that handle token refresh requests through the Comlink communication channel.

The implementation detects 401 errors and automatically requests a new token from the parent window when running in Dashboard mode. It also exposes the necessary types and utilities for token management.

### What to review

- The new `ComlinkTokenRefreshProvider` component and its integration with `AuthBoundary`
- The token refresh mechanism that handles 401 errors automatically
- The timeout handling for token refresh requests
- The exported types and utilities for Comlink communication

### Testing

- Added comprehensive tests for the token refresh functionality
- Tested different scenarios: successful token refresh, timeouts, error handling
- Verified behavior in both Dashboard and standalone modes

### DIFF
--- a/packages/react/src/components/auth/LoginError.tsx
+++ b/packages/react/src/components/auth/LoginError.tsx
@@ -41,10 +41,7 @@ export function LoginError({error, resetErrorBoundary}: LoginErrorProps): React.
   }, [logout, resetErrorBoundary])
 
   useEffect(() => {
-    if (
-      // (authState.type === AuthStateType.ERROR || authState.type === AuthStateType.LOGGED_IN) &&
-      error instanceof ClientError
-    ) {
+    if (error instanceof ClientError) {
       if (error.statusCode === 401) {
         handleRetry()
       } else if (error.statusCode === 404) {
--- a/packages/react/src/context/ComlinkTokenRefresh.tsx
+++ b/packages/react/src/context/ComlinkTokenRefresh.tsx
@@ -9,14 +9,7 @@ import {
   setAuthToken,
   type WindowMessage,
 } from '@sanity/sdk'
-import React, {
-  createContext,
-  type PropsWithChildren,
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-} from 'react'
+import React, {type PropsWithChildren, useCallback, useEffect, useMemo, useRef} from 'react'
 
 import {useAuthState} from '../hooks/auth/useAuthState'
 import {useWindowConnection} from '../hooks/comlink/useWindowConnection'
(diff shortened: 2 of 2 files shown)


joshuaellis:
related issue - DASH-390

ricokahler (commented):
i think it works but i'd like to see if we could move some of this logic into core and utilize our store architecture to propagate events instead of relying on react and useEffect. this might involve moving some of the window connection logic into store as well but i think that would be worth the effort if you'd like to explore that.

packages/react/src/components/auth/AuthBoundary.tsx
```diff
@@ -145,6 +151,7 @@ function AuthSwitch({
 
   const isLoggedOut = authState.type === AuthStateType.LOGGED_OUT && !authState.isDestroyingSession
   const loginUrl = useLoginUrl()
+  useComlinkTokenRefresh() // Initialize the comlink token refresh context for Dashboard Comlink
```
ricokahler, line 154: i see that you're calling this hook but this hook implementation doesn't seem to have any side effects.it looks like it just gives access to some functions from the context. i think the provider is the one doing the side effects and this line may not be necessary.

packages/react/src/context/ComlinkTokenRefresh.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }),
+    [],
+  )
+
+  return (
+    <ComlinkTokenRefreshContext.Provider value={contextValue}>
+      {children}
+    </ComlinkTokenRefreshContext.Provider>
+  )
+}
```
ricokahler, line 154: i don't think you need a no-op provider. i think you could omit the hook entirely with the current implementation and only continue to conditionally wrap with the DashboardTokenRefresh provider component and it should still work as expected

packages/react/src/components/auth/LoginError.tsx
```diff
@@ -33,11 +41,14 @@ export function LoginError({error, resetErrorBoundary}: LoginErrorProps): React.
   }, [logout, resetErrorBoundary])
 
   useEffect(() => {
-    if (authState.type === AuthStateType.ERROR && authState.error instanceof ClientError) {
-      if (authState.error.statusCode === 401) {
+    if (
+      // (authState.type === AuthStateType.ERROR || authState.type === AuthStateType.LOGGED_IN) &&
+      error instanceof ClientError
+    ) {
+      if (error.statusCode === 401) {
```
ricokahler, line 48: you have some commented out code here. was this change intended?

ricokahler (approved):

PR #503, opened 2025-05-05
fix(core): use raw perspective for functions that coalesce draft/published

cngonzalez:
### Description
In their respective stores, reviews and projections coalesce draft and query results and return a "status" object that includes params for when the draft was last edited and when the published version was last edited, so people could indicate the status of the document. However, they were also using the `drafts` perspective, which does the coalescing in the query result so everything was mistakenly "published" from the status perspective.

This restores the raw perspective. 

### What to review
It's a larger question about what we want to do about "status". This may only get more complex once we introduce perspectives. Are we going to return the version document and the published document?

Should we just use the `drafts` perspective and get the data some other way?

### DIFF
--- a/packages/core/src/preview/util.ts
+++ b/packages/core/src/preview/util.ts
@@ -2,7 +2,7 @@ import {getEnv} from '../utils/getEnv'
 import {type PreviewValue, type ValuePending} from './previewStore'
 
 export const PREVIEW_TAG = 'preview'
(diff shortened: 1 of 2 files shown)


cngonzalez:
Tagging you @ricokahler just for some of the above questions

binoy14 (approved):

PR #504, opened 2025-05-06
feat!: make all comlink hooks suspend

cngonzalez:
### Description
This PR refactors the code used for connecting with the Dashboard to Suspend to better follow our existing practices.

This required several architectural changes:
1. Make a subscribable state source out of retrieving and releasing nodes, rather than doing them independently
2. Create a StateSourceHook out of that action to be used in the React package
3. Update all consumers of that hook to do away with "connected" states
4. Update core functions that used the getOrCreate or release functions to use the state source instead. 

### What to review
Most of the changes are in Core -- all of the React changes are just downstream to deal with the new state of things. I have most uncertainty around the RxJS methods but please do advise if you see anything else.

### Testing
Most utils are at 100% test coverage. Unfortunately this is a bit involved to test manually:

1. Pull down this branch
2. Run it locally (You may have to remove project `ezwd8xes` from the config in App.tsx in Kitchensink -- otherwise there will be an error in Core).
3. Go to https://core.sanity.io/@oSyH1iET5/application?dev=http://localhost:3333/document-dashboard-interactions and play around

Here is a GIF as well:

### DIFF
--- a/packages/react/guides/0-Migration-Guide.md
+++ b/packages/react/guides/0-Migration-Guide.md
@@ -2,6 +2,161 @@
 title: Migration guide
 ---
 
+## Migrating to @sanity/sdk-react@1.0.0
+
+### Breaking Changes
+
+1. `useManageFavorite`, `useNavigateToStudioDocument`, and `useRecordDocumentHistoryEvent` now all require a Suspense boundary.
+
+**Before:**
+
+```typescript
+function MyDocumentAction(props: DocumentActionProps) {
+  const {documentId, documentType, resourceId} = props
+  const {favorite, unfavorite, isFavorited, isConnected} = useManageFavorite({
+    documentId,
+    documentType,
+    resourceId
+  })
+
+  return (
+    <Button
+      disabled={!isConnected}
+      onClick={() => isFavorited ? unfavorite() : favorite()}
+      text={isFavorited ? 'Remove from favorites' : 'Add to favorites'}
+    />
+  )
+}
+```
+
+**After:**
+
+```typescript
+function FavoriteButton(props: DocumentActionProps) {
+  const {documentId, documentType, resourceId} = props
+  const {favorite, unfavorite, isFavorited, isConnected} = useManageFavorite({
+    documentId,
+    documentType,
+    resourceId
+  })
+
+  return (
+    <Button
+      disabled={!isConnected}
+      onClick={() => isFavorited ? unfavorite() : favorite()}
+      text={isFavorited ? 'Remove from favorites' : 'Add to favorites'}
+    />
+  )
+}
+
+// Wrap the component with Suspense since the hook may suspend
+function MyDocumentAction(props: DocumentActionProps) {
+  return (
+    <Suspense fallback={<Button text="Loading..." disabled />}>
+      <FavoriteButton {...props} />
+    </Suspense>
+  )
+}
+```
+
+---
+
+**The following hooks now also suspend and must be wrapped in `<Suspense>`:**
+
+### `useNavigateToStudioDocument`
+
+**Before:**
+
+```typescript
+function NavigateButton({documentHandle}: {documentHandle: DocumentHandle}) {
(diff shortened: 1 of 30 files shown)


packages/react/guides/0-Migration-Guide.md
```diff
@@ -2,6 +2,161 @@
 title: Migration guide
 ---
 
+## Migrating to @sanity/sdk-react@1.0.0
```
cngonzalez, line 5: I'm unsure if we want a new file for this.

cngonzalez:
technically this also resolves SDK-415 since the hook will suspend if the node connection is not available and does not provide a fallback state or retry

packages/react/guides/0-Migration-Guide.md
```diff
@@ -2,6 +2,161 @@
...
+  )
+}
+```
+
+**After:**
+
+```typescript
+function FavoriteButton(props: DocumentActionProps) {
+  const {documentId, documentType, resourceId} = props
+  const {favorite, unfavorite, isFavorited, isConnected} = useManageFavorite({
```
ricokahler, line 37: 💣

```suggestion
  const {favorite, unfavorite, isFavorited} = useManageFavorite({
```

packages/react/guides/0-Migration-Guide.md
```diff
@@ -2,6 +2,161 @@
...
+    />
+  )
+}
+```
+
+**After:**
+
+```typescript
+function NavigateButton({documentHandle}: {documentHandle: DocumentHandle}) {
+  const {navigateToStudioDocument, isConnected} = useNavigateToStudioDocument(documentHandle)
```
ricokahler, line 87:
```suggestion
  const {navigateToStudioDocument} = useNavigateToStudioDocument(documentHandle)
```

packages/react/guides/0-Migration-Guide.md
```diff
@@ -2,6 +2,161 @@
 title: Migration guide
 ---
 
+## Migrating to @sanity/sdk-react@1.0.0
+
+### Breaking Changes
+
+1. `useManageFavorite`, `useNavigateToStudioDocument`, and `useRecordDocumentHistoryEvent` now all require a Suspense boundary.
```
ricokahler, line 9: i'm not sure if we necessarily need to say these require a suspense boundary but rather that they all suspend now. for example, there are cases where it may be better to rely a parent suspense boundary where potentially a `useProjection` would also suspend too and both of those hooks could share the same loading fallback. additionally the SDK now forces in one top-level boundary

ricokahler (commented):
i didn't have time to take a very in-depth look but this is a very solid improvement and should go in asap!

i'd say this is still missing another layer to 100% suspense compliance and that's primarily because `useManageFavorite` does not suspend while it's resolving `isFavorited`. to align with suspense, it would be better if `useManageFavorite` did not resume rendering until it has the correct value of `isFavorited`. looking at the code now, it looks like `isFavorited` defaults to false. this could cause issues if the user does something conditional depending on this state

cngonzalez:
> i didn't have time to take a very in-depth look but this is a very solid improvement and should go in asap!
> 
> i'd say this is still missing another layer to 100% suspense compliance and that's primarily because `useManageFavorite` does not suspend while it's resolving `isFavorited`. to align with suspense, it would be better if `useManageFavorite` did not resume rendering until it has the correct value of `isFavorited`. looking at the code now, it looks like `isFavorited` defaults to false. this could cause issues if the user does something conditional depending on this state
> 
> take a look at the migration guide, i think some `isConnected`s were left in

Thank you! Sorry for the rush (this was also done in a bit of a rush), and it's a good note on suspending for the resolution of the isFavorited state. Ideally that should be polling and do some other cool things but I think additional improvements there won't be breaking, at least. 

Added the migration guide changes in a final commit if you'd like to just review that.

ricokahler (approved):
boop

PR #512, opened 2025-05-07
chore: eslint ignore auto-generated typegen file

ryanbonial:
### Description

Added `sanity.types.ts` to the ESLint ignore list in the kitchensink-react app. This file is generated by Sanity TypeGen and should not be linted.

### What to review

Check that the ESLint configuration properly ignores the `sanity.types.ts` file, which is now listed under the "Ignore files for Sanity TypeGen" comment.

### Testing

Verified that ESLint no longer attempts to lint the `sanity.types.ts` file when running linting commands.


ricokahler (approved):
ty!

PR #523, opened 2025-05-08
fix(docs): use correct hook in TS doc

colepeters:
### Description

Fixes a hook used in the typescript doc

### What to review

- Correct hook used?

### Testing

N/A


ricokahler (approved):
tysm!

PR #527, opened 2025-05-12
fix!: change `status` to `_status` in projection and preview

ryanbonial:
### Description

Renamed `status` to `_status` in preview and projection results to prevent collisions with user-defined fields and follow the convention of using underscore prefix for system attributes. This change affects the `PreviewValue` interface and both projection and preview results.

Added migration documentation for this breaking change in the migration guide.

### What to review

- Verify that all references to the `status` field in preview and projection results have been updated to `_status`
- Check the migration guide to ensure it clearly explains the change and provides appropriate examples
- Confirm that all tests have been updated to reflect this change

### Testing

Updated all relevant tests to use `_status` instead of `status`. Tests now verify that preview and projection results correctly include the renamed field.


cngonzalez (approved):
Makes sense!

PR #539, opened 2025-05-20
chore: clarify the standalone auth is experimental

ryanbonial:
### Description

This PR clarifies the authentication documentation by:
- Specifying that standalone applications are only supported in "very limited cases"
- Adding "When should you use this?" guidance for each authentication type
- Marking the Standalone authentication mode as "Experimental"
- Providing more explicit warnings about limitations of standalone authentication

These changes help developers better understand which authentication approach is appropriate for their specific use case and set proper expectations about standalone authentication support.

### What to review

- Accuracy of the added guidance for each authentication type
- Clarity of the warnings about standalone authentication limitations
- Appropriateness of marking standalone authentication as "experimental"

### Testing

This change only affects documentation, so no functional testing is required. Please review the content for technical accuracy and clarity.


packages/core/guides/AuthenticationGuide.md
```diff
@@ -17,12 +17,21 @@ The primary interactive authentication flow involves redirecting the user to `sa
...
 
-- **Standalone:** Supports manually provided tokens (stable for backends,
-  requires care for stamped tokens on frontends). ⚠️ The built-in web login flow (`sanity.io/login`) has
-  significant limitations for apps on custom domains due to origin restrictions.
+  - **When should you use this?** This is only for using the SDK within the Sanity Studio application.
+
+- **Standalone (Experimental):** Supports manually provided tokens (stable for backends,
+  requires care for stamped tokens on frontends).
+
+  - **When should you use this?** For most developers, this is not the best option. It is only for using the SDK within a standalone application that is not running in the Sanity Dashboard and is not part of the Studio. It likely requires you to write custom code to handle the login flow and the callback.
```
colepeters, line 30: Might be good to know why this is not the best option, and why we're publishing the information around this if it's not!

ryanbonial: It is mainly not the best because it is not the *easiest* because "It likely requires you to write custom code to handle the login flow and the callback."

kristoferjoseph (approved):

PR #561, opened 2025-06-12
test(e2e): add dataset setup and teardown functions for use in CI

cngonzalez:
### Description

Added dataset setup and teardown functionality for E2E tests, with utilities for managing test datasets. This PR:

- Creates a dataset setup script that creates primary and secondary datasets if they don't exist
- Adds a dataset teardown script that cleans up test datasets in CI environments (users may want to have their datasets stick around for debugging purposes. we can enforce this differently if things get messy)
- Implements helper utilities like `sanityIdify` to create valid Sanity dataset names
- Adds a timer utility with visual feedback for long-running operations
- Fixes the CI environment variable type to be a boolean instead of string
- Normalizes some naming of setup / teardown functions

### What to review

- The dataset setup and teardown scripts in `packages/@repo/e2e/src/setup/datasets.setup.ts` and `packages/@repo/e2e/src/teardown/datasets.teardown.ts`
- The helper utilities for sanitizing dataset names and timing operations
- The type change for the CI environment variable from string to boolean
- The dependency changes in package.json files

### Testing
You can test this locally by pulling down the branch and updating the following CI variables (please look in 1pass if there's some you don't have):
```
SDK_E2E_DATASET_0=whatever-name
SDK_E2E_DATASET_1=whatever-name-2

CI=true
SDK_E2E_USER_ID=
SDK_E2E_USER_PASSWORD=
RECAPTCHA_E2E_STAGING_KEY=
```
and then running `pnpm test:e2e`

You should see console logging about creating and destroying datasets. When you go to https://www.sanity.work/organizations/oFvj4MZWQ/project/3j6vt2rg/datasets

You shouldn't see the datasets after the tests have run.


packages/@repo/e2e/src/setup/datasets.setup.ts
```diff
@@ -0,0 +1,35 @@
...
+import {sanityIdify} from '../helpers/sanityIdify'
+import {startTimer} from '../helpers/timer'
+
+const env = getE2EEnv()
+
+setup('setup test datasets', async () => {
+  const primaryDataset = sanityIdify(env.SDK_E2E_DATASET_0)
+  const secondaryDataset = sanityIdify(env.SDK_E2E_DATASET_1)
+
+  const client = getClient('primary')
```
colepeters, line 14: I'm probably misunderstanding something here, but would `'primary'` ever exist on setting up tests? And also where does the string `'primary'` come from?

colepeters (commented):
Nice!! Just minor questions!

packages/@repo/e2e/src/setup/datasets.setup.ts
cngonzalez: leftover from a older, worse way of identifying clients. didn't actually have any effect because that dataset name is never used, but fixed now!

packages/@repo/e2e/src/helpers/sanityIdify.ts
```diff
@@ -0,0 +1,9 @@
+import {deburr} from 'lodash-es'
+
+// we need to make a valid name for a sanity dataset from inputs
+// (people may have odd characters in branch names, etc)
+export function sanityIdify(input: string): string {
```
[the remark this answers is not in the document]
cngonzalez: stolen shamelessly from monorepo, i've called it something clearer now

packages/@repo/e2e/src/scripts/cleanup-datasets.ts
```diff
@@ -0,0 +1,46 @@
...
+import {startTimer} from '../helpers/timer'
+
+const env = getE2EEnv()
+
+// must be run as a separate script to avoid race conditions with the tests
+async function cleanupDatasets() {
+  const primaryDataset = sanitizeDatasetName(env.SDK_E2E_DATASET_0)
+  const secondaryDataset = sanitizeDatasetName(env.SDK_E2E_DATASET_1)
+
+  const client = getClient('primary')
```
colepeters, line 16: Is this also a holdover, or does 'primary' refer to a dataset in this case?

colepeters (approved):

PR #568, opened 2025-06-24
test(e2e): run e2e test suite in Dashboard

cngonzalez:
### Description
This PR adds a number of helpers and additional logic to ensure that e2e tests can work both in a standalone context as well as in the Dashboard. It adds an additional Playwright project that runs its tests in the Dashboard, as well as an additional fixture to ensure we can target elements both inside and outside of an iframe.

### What to review
- I've removed the "local dev" method to authenticate e2e tests using a user token. Since local devs will also run the Dashboard test project, they would not be able to effectively sign into Dashboard with just that token. I apologize for making this more complicated.
- Does the pageContext fixture make sense?
- Any other specific logic that might be obscure to future devs?

### Testing
`dashboard-chromium` runs tests within the dashboard. I'll add a test for something like `useNavigateToStudioDocument` soon.

### DIFF
--- a/apps/kitchensink-react/e2e/document-list.spec.ts
+++ b/apps/kitchensink-react/e2e/document-list.spec.ts
@@ -1,7 +1,12 @@
 import {expect, test} from '@repo/e2e'
 
 test.describe('Document List', () => {
-  test('can list and update author documents', async ({page, getClient, createDocuments}) => {
+  test('can list and update author documents', async ({
+    page,
+    getClient,
+    createDocuments,
+    getPageContext,
+  }) => {
     const client = getClient()
 
     // Create 10 author documents in a single transaction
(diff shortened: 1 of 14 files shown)


apps/kitchensink-react/e2e/document-list.spec.ts
```diff
@@ -11,30 +16,35 @@ test.describe('Document List', () => {
         name: `Test Author ${i}`,
         biography: `This is a test biography for author ${i}`,
       })),
+      {asDraft: false},
```
cngonzalez, line 19: This is an attempt to resolve a bit of flakiness, anecdotally I found that published documents were "visible" to queries a bit faster.

apps/kitchensink-react/e2e/document-list.spec.ts
```diff
@@ -11,30 +16,35 @@ test.describe('Document List', () => {
...
     )
 
     // Query for the most recent document by _updatedAt
     // (it should be at the top of the list and thus visible in the viewport for the test)
     const mostRecentDoc = await client.fetch(
-      `*[_type == "author"] | order(_updatedAt desc)[0]{_id}`,
+      `*[_id in $documentIds] | order(_updatedAt desc)[0]{_id}`,
```
cngonzalez, line 25: also more precision to avoid flake

cngonzalez:
> Bit of a general question before I get further into this! What's the reason for running tests in both Dashboard and Standalone mode? My impression is that Dashboard mode is what we want folks to be using right now, so I'm just wondering if testing in Standalone mode is a complication we want to account for at the moment or not…?

It's a good question! I suppose I see these e2e tests more for our internal devs to test against APIs and backend services. With the exception of a few things (intents / various comlink services, and authentication) the Dashboard is (or should be) pretty immaterial.

So while external apps should be built in Dashboard, and we should certainly test against it (as hopefully this PR helps us do), I don't know how important it is for ensuring a combination of SDK hooks work together well. I could be wrong about this, though.

colepeters:
That makes sense to me, thanks @cngonzalez! Back to reviewing now!

apps/kitchensink-react/e2e/document-list.spec.ts
colepeters: I'm curious what the flakiness was here?

cngonzalez: In local setups (not CI), the runners for all the jobs (webkit, chrome, dashboard-chrome, etc) run in parallel against the same dataset. one job might finish up and the document that it created might get deleted, leading to errors. so this is really just best practice, we should only work with the documents _our_ job has created.

packages/@repo/e2e/src/helpers/pageContext.ts
```diff
@@ -0,0 +1,67 @@
+import {type FrameLocator, type Locator, type Page} from '@playwright/test'
+
+/**
+ * Unified interface for detecting elements in pages whether in iframe or not
+ * (this might get lengthy with time but hopefully what we've provided here is sufficient)
+ */
```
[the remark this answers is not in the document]
cngonzalez: good feedback! added now :)

packages/@repo/e2e/src/helpers/pageContext.ts
```diff
@@ -0,0 +1,67 @@
...
+    role: Parameters<Page['getByRole']>[0],
+    options?: Parameters<Page['getByRole']>[1],
+  ) => Locator
+  /** Get a locator by any selector */
+  locator: (selector: string) => Locator
+  /** Whether we're running in dashboard context */
+  isDashboard: boolean
+  /** The underlying page or frame */
+  context: Page | FrameLocator
+}
```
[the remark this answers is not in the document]
cngonzalez: good feedback! added now

ricokahler (approved):

PR #580, opened 2025-07-16
feat(core): allow a userId option for the user store

ryanbonial:
### Description

Added functionality to fetch and resolve individual users by ID in the users store. This includes support for both global-scoped and project-scoped user IDs, with appropriate API handling for each type. Also added a new `test:coverage:open` script to easily view coverage reports.

### What to review

- New exports in `packages/core/src/_exports/index.ts` for user-related functions and types
- Implementation of `getUserState` and `resolveUser` functions in `usersStore.ts`
- Updated types to support fetching individual users
- Tests for the new user fetching functionality
- Configuration updates to support the new coverage report script

### Testing

Added comprehensive tests for the new user fetching functionality, including tests for both global-scoped and project-scoped user IDs. The tests verify that the correct API endpoints are called and that the response is properly transformed.

### Fun gif
chosen

### DIFF
--- a/packages/core/src/users/usersConstants.ts
+++ b/packages/core/src/users/usersConstants.ts
@@ -1,5 +1,5 @@
 // NOTE: currently this API is only available on vX
 export const API_VERSION = 'vX'
-export const PROJECT_API_VERSION = 'v2021-06-07'
+export const PROJECT_API_VERSION = '2025-07-18'
 export const USERS_STATE_CLEAR_DELAY = 5000
 export const DEFAULT_USERS_BATCH_SIZE = 100
--- a/packages/core/src/users/usersConstants.ts
+++ b/packages/core/src/users/usersConstants.ts
@@ -1,4 +1,5 @@
 // NOTE: currently this API is only available on vX
 export const API_VERSION = 'vX'
+export const PROJECT_API_VERSION = 'v2021-06-07'
 export const USERS_STATE_CLEAR_DELAY = 5000
 export const DEFAULT_USERS_BATCH_SIZE = 100
--- a/packages/core/src/users/usersStore.ts
+++ b/packages/core/src/users/usersStore.ts
@@ -43,7 +43,9 @@ import {
   updateLastLoadMoreRequest,
 } from './reducers'
 import {
+  type GetUserOptions,
   type GetUsersOptions,
+  type ResolveUserOptions,
   type ResolveUsersOptions,
   type SanityUser,
   type SanityUserResponse,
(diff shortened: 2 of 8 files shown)


packages/core/src/users/usersStore.ts
```diff
@@ -313,3 +411,30 @@ export const loadMoreUsers = bindActionGlobally(
     return await promise
   },
 )
+
+/**
+ * @beta
+ */
+export const getUserState = bindActionGlobally(
```
ryanbonial, line 418: This is a convenience function. Ideally `getUsersState` (plural) wouldn't be overloaded with a single userId option but that would take a bit more of a refactor.

packages/core/src/users/usersConstants.ts
```diff
@@ -1,4 +1,5 @@
 // NOTE: currently this API is only available on vX
 export const API_VERSION = 'vX'
+export const PROJECT_API_VERSION = 'v2021-06-07'
```
cngonzalez, line 3: nit: why such an early version?

packages/core/src/users/usersStore.ts
```diff
@@ -100,9 +104,98 @@ const listenForLoadMoreAndFetch = ({state, instance}: StoreContext<UsersStoreSta
...
           switchMap((e) => {
             if (!e.added) return EMPTY
-            const {batchSize, ...options} = parseUsersKey(group$.key)
+            const {userId, batchSize, ...options} = parseUsersKey(group$.key)
 
-            const projectId = options.projectId ?? instance.config.projectId
+            if (userId) {
+              // In the future we will be able to fetch a user from the global API using the resourceUserId,
+              // but for now we need to use the project subdomain to fetch a user
+              if (userId.startsWith('p')) {
```
cngonzalez, line 112: Should we make it clear in this comment that a userId starting with p is a project user?

packages/core/src/users/usersStore.ts
```diff
@@ -144,6 +237,22 @@ const listenForLoadMoreAndFetch = ({state, instance}: StoreContext<UsersStoreSta
               filter((cursor) => cursor !== null),
             )
 
+            if (userId) {
+              return combineLatest([resource$, client$]).pipe(
+                switchMap(([resource, client]) =>
+                  client.observable.request<SanityUserResponse>({
```
cngonzalez, line 243: I'm sorry, I'm not super familiar with the access API or this store. Why is it being called here as well as in the switch map above? Is this meant to be an initial request?

cngonzalez (commented):
One (possibly silly) question and one suggestion. We should probably consider an e2e test at some point too.

packages/core/src/users/usersConstants.ts
ryanbonial: Good call. We have this version a lot in the project, I'll open a issue to update them

packages/core/src/users/usersStore.ts
ryanbonial: Great catch. This was leftover copy/paste dead code.

ryanbonial:
Thanks for the review. All comments have been addressed.

PR #585, opened 2025-07-22
feat(core): add defineIntent

cngonzalez:
### Description
FIXES SDK-525

This PR adds a `defineIntent` function. Like `defineType`, it's mostly there as a type helper for our users, but should also hopefully provider helpful validation and error states when intents are declared.

I also plan on using this in the CLI, to be used if people declare intents as bare objects (we can use this to validate / normalize).

### What to review
Mostly should be self-explanatory. Two questions:
1. are we happy where this is located (in core) and its place in the file structure (in utils?)
2. does the filters system make sense? Any gotchas there?

### Testing
100% test coverage, mostly for the error scenarios.

### DIFF
--- a/packages/core/src/utils/defineIntent.ts
+++ b/packages/core/src/utils/defineIntent.ts
@@ -1,0 +1,249 @@
+/**
+ * Filter criteria for intent matching. Can be combined to create more specific intents.
+ *
+ * @example
+ * ```typescript
+ * // matches only geopoints in the travel-project project, production dataset
+ * const filter: IntentFilter = {
+ *   projectId: 'travel-project',
+ *   dataset: 'production',
+ *   types: ['geopoint']
+ * }
+ *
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/defineIntent.ts
```diff
@@ -0,0 +1,249 @@
...
+ * }
+ *
+ * const intent = await createAsyncIntent()
+ * export default intent
+ * ```
+ *
+ * @public
+ */
+export function defineIntent(intent: Intent): Intent {
+  // Validate required fields
```
ryanbonial, line 146: [non-blocking] It would be nice to have the PR comment added in the code "This will be used this in the CLI, if people declare intents as bare objects (we can use this to validate / normalize)"

ryanbonial (approved):
The type and validation look good.

ryanbonial (approved):

PR #587, opened 2025-07-22
feat(react-internal): create react-internal package

cngonzalez:
### Description
Creates a new package specifically for React hooks to be used by internal teams to be published to a restricted org-only package. No content and most of the configuration is doubled from `@sanity/sdk-react`. 

This will be initially used for intent resolution logic to be used in Dashboard.

### What to review
Do we like this here or do we want a private repo? Are we okay with the naming? Does the release configuration look correct?

### Testing
This should hopefully build. Releasing it is another story.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,5 @@
 {
   "packages/core": "2.1.0",
-  "packages/react": "2.1.0"
+  "packages/react": "2.1.0",
+  "packages/react-internal": "0.1.0-alpha.0"
 }
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -56,6 +56,15 @@ const baseConfig = {
       entry: ['package.bundle.ts'],
       ignoreDependencies: ['@sanity/browserslist-config', 'react-compiler-runtime'],
     },
+    'packages/react-internal': {
+      typescript: {
+        config: 'tsconfig.settings.json',
+      },
+      project,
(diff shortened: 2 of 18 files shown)


binoy14:
Naming question, instead of a new `react-internal` package. Why not have an import that is `@sanity/sdk-react/_internal` similar to how studio does it?

binoy14:
> a bit more freedom around things like breaking changes
this repo is configured to have the same version for all packages and it will be annoying to configure this. In studio anything marked as internal does not have to follow semver, we can change at will.

> publishing to an org-restricted package.
If this is going to be private then we have more issues around provenance and such which we had to deal with in the past. Not the end of the world but it is nice to not deal with it. 

To me it sounds like the internal package can also be outside of this repo. 

No strong opinions but just something to thing about.

cngonzalez:
> > a bit more freedom around things like breaking changes
> > this repo is configured to have the same version for all packages and it will be annoying to configure this. In studio anything marked as internal does not have to follow semver, we can change at will.
> 
> > publishing to an org-restricted package.
> > If this is going to be private then we have more issues around provenance and such which we had to deal with in the past. Not the end of the world but it is nice to not deal with it.
> 
> To me it sounds like the internal package can also be outside of this repo.
> 
> No strong opinions but just something to thing about.

Hmm then maybe the most convenient thing is to do this under an `_internal` path. I'll close this for now unless anyone is very gung-ho about this being its own package.

PR #598, opened 2025-08-06
fix(react): find correct instance for hooks that use config

cngonzalez:
### Description
There was some crossed wires / oversight in a few hooks, so passing in a `documentHandle` with a specific `projectId` or `dataset` wouldn't necessarily pull up the correct `SanityInstance`.

This PR fixes that and adds those hooks to the MultiResourceRoute and an e2e test. 

### What to review
The actual important changes are in the react package -- it's a 2 line change. The rest is just future-proofing :)

### Testing
Added e2e tests/


ryanbonial (approved):
🚀

PR #606, opened 2025-08-27
fix: support params in use document projection

SimeonGriggs:
### Description

Update `useDocumentProjection` to use params

### What to review

This was coded by GPT-5, sorry.

### Testing

I asked GPT-5 to add tests


cngonzalez:
This is great! I'm in the middle of another refactor of the projectionStore that should hopefully simplify the store (no more overlays of different versions, statuses fetched separately) so i might wait for that one to go through before revising this.

(for example: we should likely put tests in core rather than in react but i can handle that after the other PR)

PR #607, opened 2025-08-27
feat(projects): add organizationId and includeMembers parameters to useProjects hook

ryanbonial:
### Description

- Add optional organizationId parameter for filtering projects by organization
- Add optional includeMembers parameter
- Implement parameter-based caching with unique cache keys

### What to review

The changes to the projects fetcher store in `projects.ts` which now supports filtering by organization and toggling member inclusion. The cache key generation now incorporates these parameters to ensure proper caching behavior.

The updated type definitions in `useProjects.ts` that reflect the new parameter options and return types.

### Testing

Added test coverage for the parameter passing behavior. The test in `projects.test.ts` now verifies that the correct parameters are passed to the list method.

### Fun gif
organization

### DIFF
--- a/packages/core/src/projects/projects.test.ts
+++ b/packages/core/src/projects/projects.test.ts
@@ -36,7 +36,7 @@ describe('projects', () => {
 
     const result = await resolveProjects(instance)
     expect(result).toEqual(projects)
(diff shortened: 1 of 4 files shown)


ryanbonial:
Ready for another review

binoy14 (approved):
makes sense to me, thanks

PR #609, opened 2025-09-02
fix(core): add perspective to query cache key to avoid collisions

cngonzalez:
### Description
There was a subtle bug that could happen if you used a nested `ResourceProvider` with a perspective (or any other way of implicitly getting a perspective) that pointed to the same project and dataset as another `ResourceProvider`, wherein you might get the wrong query result. You can see this if you run this branch  and check the new route in Kitchensink without the queryStore fix, as below:

Every query store is instanced to a project and dataset, and it caches the results of a query. If you were to make the exact same query in a different provider with a different perspective, your "new" perspective would not be saved in the query key, and you would be served the old cached query result with from the "wrong" perspective, because neither query saved their perspective as part of their query key.

(this wasn't a problem with something like `useDocument({...documentHandle, perspective})` because all options got memoized into the key)

This PR rectifies this by forcing every query key to have a perspective. It also adds tests and e2e tests. New results below:

### What to review
The only significant logic change is in `queryStore.ts`. Everything else is testing / future-proofing.

- The new `PerspectivesRoute` component that demonstrates nested `ResourceProvider` components with different perspectives
- The updated query store implementation that now properly separates cache entries by perspective
- The new e2e test that verifies published panels don't show draft content

### Testing

Added a comprehensive e2e test that:
- Creates a published author document
- Creates a draft overlay with modified content
- Verifies that the drafts panel shows the draft content
- Verifies that the published panel shows only the published content

### DIFF
--- a/apps/kitchensink-react/src/routes/releases/ReleasesRoute.tsx
+++ b/apps/kitchensink-react/src/routes/releases/ReleasesRoute.tsx
@@ -114,16 +114,16 @@ export function ReleasesRoute(): JSX.Element {
     () => ({...selectedDocument, perspective: selectedPerspective.perspective}),
     [selectedDocument, selectedPerspective],
   )
-
+  const documentProjectionOptions = useMemo(
+    () => ({
+      ...documentOptions,
+      projection: `{name, "bestFriend": bestFriend->name}` as `{${string}}`,
+    }),
+    [documentOptions],
+  )
   const documentResult = useDocument(documentOptions)
   const previewResult = useDocumentPreview(documentOptions)
-  const projectionResult = useDocumentProjection({
-    ...documentOptions,
-    projection: `{
-      name,
-      "bestFriend": bestFriend->name
-    }`,
-  })
+  const projectionResult = useDocumentProjection(documentProjectionOptions)
 
   return (
     <Box padding={4}>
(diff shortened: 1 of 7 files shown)


cngonzalez:
* **#609** <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/609?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

apps/kitchensink-react/src/routes/releases/ReleasesRoute.tsx
```diff
@@ -114,16 +114,17 @@ export function ReleasesRoute(): JSX.Element {
     () => ({...selectedDocument, perspective: selectedPerspective.perspective}),
     [selectedDocument, selectedPerspective],
   )
-
+  const documentProjectionOptions = useMemo(
```
cngonzalez, line 117: Not related, but the old implementation caused infinite loading due to recalculating the params passed to documentProjection. That might be something we want to resolve in the hook itself at a later date, but it was annoying me 😅

binoy14 (approved):
makes sense to me, thanks

PR #623, opened 2025-09-12
feat(core): allow multiple stores of same type per instance

cngonzalez:


cngonzalez:
Closing since there's a broader refactor in play.

PR #624, opened 2025-09-15
chore: merge kitchensink's Vite config into CLI config

ryanbonial:
### Description

This PR extends the Sanity CLI configuration in the kitchensink-react app to integrate with the app's Vite configuration. It adds a `vite` function that merges the app's Vite config with Sanity CLI's internal Vite config, handling plugin deduplication and ensuring React Refresh isn't injected twice. The implementation carefully merges plugins, resolve aliases, define variables, and server settings.

### What to review

- The plugin merging logic that prevents duplicates and avoids double React Refresh injection
- The resolution of the app's Vite config and proper merging with Sanity CLI's config
- The handling of different configuration sections (plugins, resolve, define, server)
- The TypeScript types imported from Vite

### Testing

This change should be tested by running the kitchensink-react app in both development and production modes to ensure that:
- The app builds and runs correctly
- Vite plugins from both configs are properly applied
- React Refresh works as expected in development mode
- No console errors related to duplicate plugins or configurations

### Fun gif
drift merge

### DIFF
--- a/apps/kitchensink-react/sanity.cli.ts
+++ b/apps/kitchensink-react/sanity.cli.ts
@@ -1,9 +1,87 @@
 import {defineCliConfig} from 'sanity/cli'
+import {
+  type ConfigEnv,
+  type Plugin,
+  type PluginOption,
+  type UserConfig,
+  type UserConfigExport,
+} from 'vite'
 
 export default defineCliConfig({
   app: {
     organizationId: 'oblZgbTFj',
     entry: './src/App.tsx',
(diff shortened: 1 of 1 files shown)


ryanbonial:
The vite config for kitchensink is still used here: https://github.com/sanity-io/sdk/pull/624/files#diff-512fc0140461586bb16eb348fedfebbe6e12ebd52e375f1cea5c8d9628945ae0R18 and also will be backwards compatable and used if we need to call vite directly

PR #627, opened 2025-09-16
fix(core): change ValidProjection to string type

cngonzalez:
### Description
The template literal type for ValidProjection was a bit too strict, pretty much always resulting in error and user confusion. As far as I can tell there is no way to declare this type to be a bit more forgiving without making something kind of messy, so I've deprecated it and changed it to a string.

I've kept the type for backwards compatibility but redeclared it to a string as well.

### What to review
Very little has changed -- I just wanted to ensure that any function that received a projection as a parameter would validate the projection.

### Testing
Existing tests should suffice.

### DIFF
--- a/packages/core/src/projection/types.ts
+++ b/packages/core/src/projection/types.ts
@@ -1,10 +1,5 @@
 /**
  * @public
- */
-export type ValidProjection = `{${string}}`
-
-/**
- * @public
  * The result of a projection query
  */
 export interface ProjectionValuePending<TValue extends object> {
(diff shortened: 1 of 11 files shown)


packages/core/src/projection/types.ts
```diff
@@ -16,8 +11,16 @@ export interface DocumentProjectionValues<TValue extends object = object> {
   [projectionHash: string]: ProjectionValuePending<TValue>
 }
 
+/**
+ * @public
+ * @deprecated
```
ryanbonial, line 16: (question) do we want to mark this as deprecated or do we want to mark the version that this goes out with as a breaking change?

cngonzalez: We can certainly use it as a reason to go up a version but I can't think of a scenario in which this might break someone's builds. The type has become more permissive so anyone who used ValidProjection explicitly shouldn't get any type errors, I don't think (maybe some linting against deprecations?)...

PR #630, opened 2025-09-22
feat: make applyDocumentActions more similar to the rest

judofyr:
### Description

`applyDocumentActions` was the only function in the `core` package which interpreted the `SanityInstance` as a tree, automatically traversing the tree to find a matching instance. This is inconsistent with the other stores (e.g. query store) which assumes that it's being called with the right instance.

In addition, this changes `applyDocumentActions` to only accept actions as an array (so we don't have to check if it's an array or not) and also it places the `actions` as part of option. This is consistent with e.g. `getQueryState` which has a required `query` field in its options. This now means that all functions which are wrapped in `bindActionByDataset` are now consistently getting passed an object as their first argument. This is going to be key so that we later can modify this function to consider the options (instead of the instance).

There should be no breaking changes in the React package.

### DIFF
--- a/packages/core/src/document/applyDocumentActions.test.ts
+++ b/packages/core/src/document/applyDocumentActions.test.ts
@@ -72,7 +72,8 @@ describe('applyDocumentActions', () => {
     }
 
     // Call applyDocumentActions with a fixed transactionId for reproducibility.
-    const applyPromise = applyDocumentActions(instance, action, {
+    const applyPromise = applyDocumentActions(instance, {
+      actions: [action],
       transactionId: 'txn-success',
     })
 
@@ -128,7 +129,8 @@ describe('applyDocumentActions', () => {
     }
 
     // Call applyDocumentActions with a fixed transactionId.
-    const applyPromise = applyDocumentActions(instance, action, {
+    const applyPromise = applyDocumentActions(instance, {
+      actions: [action],
       transactionId: 'txn-error',
     })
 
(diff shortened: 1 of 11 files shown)


judofyr:
* **#632** <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#631** <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#630** <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/630?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/sdk/630?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

cngonzalez (approved):
Yes I think this makes it more like other hooks. In most other hooks, the `projectId` and `dataset` are used in the React layer to find the correct instance, and core just assumes that it has the right instance passed to it to do whatever it needs to do.

Forcing actions to be an array does break the contract of this function, but it's:
1. on core, which isn't used too much on its own
2. on a function marked internal, which means users shouldn't be depending on it anyway
3. brings us closer to the syntax used elsewhere (client requires an array of actions too), and we've recently received feedback about contracts matching across product surfaces

So I'm generally not too concerned about this one.

judofyr:
Rebased once more and fixed a conflict.

PR #632, opened 2025-09-22
feat: support using the same instance for different datasets

judofyr:


cngonzalez (approved):
Appreciate the lack of breaking changes to any exported utils! Also did some manual testing in those places where there could have been some divergence (how the client is retrieved in the perspective store, for example), and saw no issues.

PR #645, opened 2025-09-26
test: removing mocking of useSanityInstance hook

judofyr:
### Description

Many of the React tests were mocking the `useSanityInstance` hook. This is unfortunate because it depends on internals (the fact that it invokes that specific hook) which we might want to change. This replaces all of these mocks by using the `ResourceProvider`. This makes the test closer to the public API and less likely to have to change if we want to change the internals.

### What to review

- Does it still test the same behavior?

### Testing

N/A.

### Fun gif

Friday


binoy14 (approved):
Looks good to me.

In future we should abstract this into `renderWithProvider` helper but that probably is out of scope for this PR

PR #648, opened 2025-10-15
fix(auth): do not clear the entire hash during login, only a token value

ryanbonial:
### Description

Improved the `getCleanedUrl` function to preserve hash fragments while only removing the token and withSid parameter. Previously, the function would remove the entire hash, but now it selectively removes only the token and withSid parameters while keeping other hash parameters intact. This provides a better user experience by maintaining URL state after authentication.

### What to review

- The updated implementation of `getCleanedUrl` in `utils.ts` which now preserves hash fragments
- The comprehensive test cases that verify different URL cleaning scenarios
- Verify that the function correctly handles:
  - URLs with token as the only hash parameter
  - URLs with token among multiple hash parameters
  - URLs with non key-value hash fragments
  - URLs with query parameters that should be preserved or removed

### Testing

Added comprehensive test cases that cover various URL scenarios:
- Removing token when it's the only hash parameter
- Preserving other hash parameters when removing token
- Handling token in the middle of multiple hash parameters
- Removing sid and url from query parameters while preserving others
- Preserving non key-value hash fragments

### Fun gif
hash

### DIFF
--- a/packages/core/src/auth/utils.ts
+++ b/packages/core/src/auth/utils.ts
@@ -126,6 +126,7 @@ export function getCleanedUrl(locationUrl: string): string {
   if (rawHash && rawHash.includes('=')) {
     const hashParams = new URLSearchParams(rawHash)
     hashParams.delete('token')
+    hashParams.delete('withSid')
     const nextHash = hashParams.toString()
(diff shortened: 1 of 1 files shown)


laurenashpole (approved):
LGTM!

PR #656, opened 2025-10-30
feat: add asset management functions and hooks

ryanbonial:
### Description

Added asset management capabilities to the SDK, including new hooks and utilities for working with assets. This PR introduces:

- Core asset management functions in `@sanity/sdk`
- React hooks in `@sanity/sdk-react` for asset operations
- A new Assets demo route in the kitchensink app

The implementation provides a complete asset management solution with support for:
- Querying assets with flexible filtering options
- Uploading images and files
- Deleting assets
- Linking Media Library assets

### What to review

- Core asset functions in `packages/core/src/assets/assets.ts`
- React hooks in `packages/react/src/hooks/assets/`
- The demo implementation in `apps/kitchensink-react/src/routes/AssetsRoute.tsx`
- Public API exports in both packages

### Testing

The implementation includes:
- Type safety with proper overloads for different asset kinds
- Comprehensive error handling
- A fully functional demo route that exercises all the new capabilities

### Fun gif
assets


packages/core/src/assets/assets.ts
```diff
@@ -0,0 +1,391 @@
...
+  },
+  fetcher: (instance) => (options) => {
+    const projectId = options.projectId ?? instance.config.projectId
+    const dataset = options.dataset ?? instance.config.dataset
+    if (!projectId || !dataset) {
+      throw new Error('A projectId and dataset are required to query assets.')
+    }
+
+    const groq = buildAssetsGroq(options)
+    return getClientState(instance, {
```
cngonzalez: Is there reason not to use the query store for this? It might be more expected behavior.

packages/core/src/assets/assets.ts
```diff
@@ -0,0 +1,391 @@
...
+}
+
+/**
+ * Build a CDN URL for an image asset from its asset ID.
+ *
+ * Expects `image-{hash}-{width}x{height}-{format}`. Throws for invalid input.
+ *
+ * @public
+ */
+export function buildImageUrlFromId(
```
cngonzalez: Would it make sense to just re-export functionality from https://github.com/sanity-io/image-url to align our logic and have a single source of truth on how this is done?

If we want to go more low-level, we could also use the exports from https://github.com/sanity-io/asset-utils

cngonzalez (changes requested):
I like having all these helper hooks! I'm not sure I fully understand why there's a need for a separate store, though. Can we treat image assets like documents?

Also added some comments on maybe using our existing tooling to keep this simple.

packages/core/src/assets/assets.ts
ryanbonial: Neat! I didn't know that package was available

packages/core/src/config/sanityConfig.ts
```diff
@@ -63,6 +63,17 @@ export interface DocumentHandle<
   documentId: string
 }
 
+/**
+ * Identifies a specific asset within a Sanity project/dataset.
+ * Includes optional `projectId`/`dataset` and the required `assetId` (asset document `_id`).
+ * Useful for cross-instance asset operations without relying on the active ResourceProvider.
```
cngonzalez: I don't think I fully understand this comment -- what is meant by cross-instance? I'm also a little wary of the mental model of "this is a document but the _id has a different parameter name"

cngonzalez (commented):
A few nits but a larger comment:

I really like the addition of the utility hooks but I'm a bit confused about the implementation of `useAssets` and the underlying `getAssetsState` et al. I've got two main questions:
1. We're using the `StateSource` type but are we actually using a state at any point? (Or the associated subscription / unsubsciption logic). I couldn't see it, but I could be wrong. Should we, to be consistent?
2. `useAssets` diverges a bit from the rest of our codebase; I think it's returning whole documents. Should it conform more closely to `useDocuments` and return handles (we might even be able to pass the buck to `getDocumentsState` and have simpler logic)? And then we can implement a `useAsset` that returns the full document that can then be a full doc? Or is that annoying in practice?

packages/core/src/assets/assets.ts
```diff
@@ -0,0 +1,371 @@
...
+ * @public
+ */
+export type ImageAssetId<TFormat extends string = string> =
+  `image-${string}-${number}x${number}-${Lowercase<TFormat>}`
+
+/**
+ * Runtime validator and type guard for image asset IDs
+ * @public
+ */
+export function isImageAssetId(value: string): value is ImageAssetId {
```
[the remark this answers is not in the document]
ryanbonial: I think that importing from the monorepo would cause a dependency loop, so I'm hesitant to add it here.

packages/core/CHANGELOG.md
```diff
@@ -2,29 +2,26 @@
```
cngonzalez: We might want to not make changes to the changelog in this PR

packages/core/src/assets/assets.ts
```diff
@@ -0,0 +1,470 @@
...
+/**
+ * Options for linking a Media Library asset to a dataset
+ * @public
+ */
+export interface LinkMediaLibraryAssetOptions<
+  TDataset extends string = string,
+  TProjectId extends string = string,
+> extends DatasetHandle<TDataset, TProjectId> {
+  assetId: string
+  mediaLibraryId: string
```
cngonzalez: Should we mark this beta? I'm a bit wary of exposing something that uses the mediaLibraryId explicitly as public, especially with upcoming sources work

cngonzalez (changes requested):
Let me know if you want to chat or if I'm misunderstanding things! Sorry to drag this out.

PR #665, opened 2025-11-12
feat: add canvas as a source for useQuery and useDispatchIntent

cngonzalez:
### Description
Adds `canvasSource` which behaves like `mediaLibrarySource`. Right now, the `source` param is only respected by `useQuery` and `useDispatchIntent`. Minimal logic was added to `useDispatchIntent` to ensure that it can dispatch intents for Canvas documents, and corresponding Kitchensink components were added to this as well.

I also corrected the resource type for the `media-library` resource type, since that's the string used basically everywhere. It doesn't change functionality, since intents don't care about resource types (only document types).

### What to review
Again, please keep me honest and make sure things are appropriately marked `beta`!

### Testing
A few tests were added to ensure 100% coverage of all logic branches. Also, Kitchensink was deployed with these changes: https://core-4owxlg8l9.sanity.build/@oblZgbTFj/application/wkyoigmzawwnnwx458zgoh46/intents

If buttons don't work, try to navigate to another app and back to the Kitchensink (it's a bug I'm working on).

### DIFF
--- a/packages/react/src/hooks/dashboard/useDispatchIntent.ts
+++ b/packages/react/src/hooks/dashboard/useDispatchIntent.ts
@@ -39,19 +39,12 @@ interface DispatchIntent {
  * Parameters for the useDispatchIntent hook
  * @beta
  */
-type UseDispatchIntentParams =
-  | {
-      action: 'edit'
-      intentId?: never
-      documentHandle: DocumentHandleWithSource
-      parameters?: Record<string, unknown>
-    }
-  | {
-      action?: never
-      intentId: string
-      documentHandle: DocumentHandleWithSource
-      parameters?: Record<string, unknown>
-    }
+interface UseDispatchIntentParams {
+  action?: 'edit'
+  intentId?: string
+  documentHandle: DocumentHandleWithSource
+  parameters?: Record<string, unknown>
+}
 
 /**
  * @beta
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/dashboard/useDispatchIntent.ts
```diff
@@ -64,7 +64,7 @@ type UseDispatchIntentParams =
  *   - `intentId` - Specific ID of the intent to dispatch. Either `action` or `intentId` is required.
  *   - `documentHandle` - The document handle containing document ID, type, and either:
  *     - `projectId` and `dataset` for traditional dataset sources, like `{documentId: '123', documentType: 'book', projectId: 'abc123', dataset: 'production'}`
- *     - `source` for media library or dataset sources, like `{documentId: '123', documentType: 'asset', source: mediaLibrarySource('ml123')}`
+ *     - `source` for media library, canvas, or dataset sources, like `{documentId: '123', documentType: 'asset', source: mediaLibrarySource('ml123')}` or `{documentId: '123', documentType: 'canvas.document', source: canvasSource('canvas123')}`
```
laurenashpole: Just for consistency, switching the document type names here:

```suggestion
 *     - `source` for media library, canvas, or dataset sources, like `{documentId: '123', documentType: 'sanity.asset', source: mediaLibrarySource('ml123')}` or `{documentId: '123', documentType: 'sanity.canvas.document', source: canvasSource('canvas123')}`
```

laurenashpole (approved):
LGTM!

PR #742, opened 2026-02-12
perf(react): stabilize config and auth hook references to prevent unnecessary re-renders

ryanbonial:
### Description

Stabilize React provider config references to eliminate unnecessary re-renders, re-subscriptions, and instance recreation in the `@sanity/sdk-react` package.

**Problem:** Several derived values (`configs`, `projectIds`, resolved studio config) were recomputed on every render, producing new object/array references even when the underlying data hadn't changed. This caused downstream providers (`ResourceProvider`, `AuthBoundary`) and hooks (`useVerifyOrgProjects`) to unnecessarily tear down and re-initialize. In `useVerifyOrgProjects`, including `error` in the `useEffect` dependency array created a subscribe → `setError` → re-render → resubscribe loop on every observable emission.

**Solution:**
- **`SDKProvider`** — Wrap `configs` and `projectIds` in `useMemo` so downstream consumers receive stable references.
- **`SanityApp`** — Destructure the studio workspace into primitives (`projectId`, `dataset`, `auth.token`) before the `useMemo`, so the memo only recomputes when actual config inputs change — not when Studio re-provides a workspace object with the same contents but a new reference. Remove the now-unused `deriveConfigFromWorkspace` helper and its `StudioWorkspaceHandle` import.
- **`useVerifyOrgProjects`** — Remove `error` from the `useEffect` dependency array to break the resubscribe loop. Simplify the early-exit reset to always call `setError(null)` (React bails out when the value is already `null`).

Closes SDK-784

### What to review

- **`packages/react/src/components/SDKProvider.tsx`** — Two new `useMemo` calls wrapping `configs` and `projectIds`. Verify the dependency arrays are correct.
- **`packages/react/src/components/SanityApp.tsx`** — Primitive destructuring of `studioWorkspace` before the `resolvedConfig` memo, removal of `deriveConfigFromWorkspace`. Confirm the resolved config shape is unchanged.
- **`packages/react/src/hooks/auth/useVerifyOrgProjects.tsx`** — `error` removed from effect deps; `setError(null)` now called unconditionally in the disabled/empty guard. Verify no edge case where error state could go stale.

### Testing

All existing tests pass (253 passed, 4 skipped). The `useVerifyOrgProjects`, `SDKProvider`, and `SanityApp` test suites all pass without modification, confirming no behavioral regressions.

A potential follow-up is adding a test that asserts subscription stability across multiple observable emissions in `useVerifyOrgProjects` to guard against future regressions of the resubscribe loop.

### Fun gif
smol tricks

### DIFF
--- a/packages/react/src/components/SDKProvider.tsx
+++ b/packages/react/src/components/SDKProvider.tsx
@@ -42,9 +42,16 @@ export function SDKProvider({
   fallback,
   ...props
 }: SDKProviderProps): ReactElement {
-  const allConfigs = Array.isArray(config) ? config : [config]
+  const allConfigs = useMemo(() => (Array.isArray(config) ? config : [config]), [config])
   const resolvedConfig = allConfigs[0]
-  const projectIds = allConfigs.map((c) => c.projectId).filter((id): id is string => !!id)
+
+  // Memoize projectIds so AuthBoundary → useVerifyOrgProjects receives a stable
+  // array reference rather than a new one on every render.
+  // Note: React Compiler should handle this automatically, but kept for explicit intent.
+  const projectIds = useMemo(
+    () => allConfigs.map((c) => c.projectId).filter((id): id is string => !!id),
+    [allConfigs],
+  )
 
   // Extract static fields so the memo below doesn't take a reference dependency
   // on `config` — inline config objects change identity on every render.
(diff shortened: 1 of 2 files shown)


ryanbonial:
Fixed on https://github.com/sanity-io/sdk/pull/824

PR #779, opened 2026-03-24
feat: add parameter to infer resources

cngonzalez:
### Description

Our Media Library and Canvas IDs aren't really exposed to users — they'd have to query an endpoint or check their network tab in Dashboard to get that information.

This PR adds a boolean prop to SanityApp: inferMediaLibraryAndCanvas. When set, SanityApp fetches the organization's resources on mount and makes them available throughout the app via the media-library and canvas resource names.

I made some choices here:

1. Opt-in. Most non-enterprise customers won't have a Media Library and may not need this. Including everyone by default would add latency for nothing.

2. React layer only. Core utilities require an explicit resource every time, and we don't have a convention for async operations at instance instantiation. If we ever add that, it'd probably live in a higher-level Node abstraction.

3. Suspends until resolved. `OrganizationResourcesProvider` uses React's `use()` to suspend its subtree while inference is in flight. The nearest `Suspense` boundary (already provided by ResourceProvider using the fallback prop) shows the loading state. This means inference adds to initial load time — the trade-off for not having to manage resource IDs manually.

### What to review

Latency. The app suspends on mount while we fetch org resources from the API. For most users this is a one-time cost at startup, and they're already waiting on auth/data anyway — but it's worth calling out. If inference fails (network error, org has no media library, etc.), we log a warning and continue without the inferred resources rather than blocking the app.

### Testing

Tests were added for the new component and function.

The e2e tests already covered Media Library read/write operations, and the kitchensink was updated for the new pattern — if those pass, this is also e2e tested.

### Fun gif

  FIXES SDK-1112


apps/kitchensink-react/src/App.tsx
```diff
@@ -25,6 +25,7 @@ export default function App(): JSX.Element {
         fallback={<Spinner />}
         config={isE2E ? e2eConfig : devConfig}
         resources={isE2E ? e2eResources : devResources}
+        inferOrganizationResourcesFrom={isE2E ? e2eOrgId : devOrgId}
```
ryanbonial: I am torn between liking a prop name that is precise about the behavior on this PR, `inferOrganizationResourcesFrom`, and wanting one that more honestly reflects the underlying value, `organizationId`, and which would leave room for future org-scoped SDK features without proliferating multiple props that all take the same ID. The tradeoff is clarity about the current implementation versus a cleaner long-term API surface. @cngonzalez do you have strong feelings on this either way?

cngonzalez: What might work (even though it's more verbose) is two props:
1. `organizationId` which could be re-used for any future-scoped utilities
2. `inferMediaLibraryAndCanvas` as a boolean, so it's still opt-in, but means that users can granularly set any future-scoped permissions going forward

ryanbonial: I'm not a big fan of having 2 props that need to be filled in so that 1 thing can function correctly. 

Would there ever be a reason that `inferMediaLibraryAndCanvas` would ever be `false`?  I don't think their is any penalty for having them always there as a named source (although many orgs won't have Media Library functionality).

Sorry to add more options, but now I am wondering if we can just do all of this for the devs with `useDashboardOrganizationId()` and then `GET /media-libraries?organizationId=<orgId>` and `GET /canvases?organizationId=<orgId>`. It would only work if the App was running in the Dashboard, but we already have quite a few things in the SDK that only work in the dashboard.

ryanbonial: ...or maybe something like

```ts
<SanityApp resources={{ default: { projectId: '...', dataset: '...' } }} fallback={<Loading />}>
  <OrganizationResourceProvider>
    <AppContent />
  </OrganizationResourceProvider>
</SanityApp>
```

and it merges the `ResourcesContext` with the Organizatoin resources. This follows with our composable and overridable patterns.

cngonzalez: > Would there ever be a reason that inferMediaLibraryAndCanvas would ever be false?

Only if we add other organization-level functionality and people want one thing but not another. 

I'm happy to have inference "just happen" if you pass an `orgId` and it be "default" behavior -- there's a lot of logic here to prevent blocking rendering. We can always add an additional parameter to opt out later.

It's just a bit opaque from a parameter perspective but nothing's perfect 😅 

> now I am wondering if we can just do all of this for the devs with useDashboardOrganizationId()

I'm unsure. I wanted to limit dependency on Dashboard since it's changing. Also waiting for the Comlink node to connect / fetch is another thing that might slow down startup, since it suspends (again, there's logic to prevent blocking render). 

You're right that it's nice not to have to explicitly pass an orgId but it's something that's in the`sanity.cli.ts` file and hopefully gets passed forward like resources anyway.

cngonzalez: i've actually changed my mind again 😵  it's a bit of a weird code smell, at least for those of us working on the SDK, to both pass an org id and also have utilities to get the org from the Dashboard. while i'm wary of the comlink dep i'll work on a way to invoke it without holding up the whole component tree. but that would mean just changing the top level parameter to `inferMediaLibraryAndCanvas`.

packages/react/src/utils/resolveOrgResources.ts
```diff
@@ -0,0 +1,59 @@
...
+ * the other from resolving.
+ */
+export async function resolveOrgResources(
+  instance: SanityInstance,
+  organizationId: string,
+): Promise<OrgResources> {
+  const client = await firstValueFrom(
+    getClientState(instance, {
+      apiVersion: API_VERSION,
+      requestTagPrefix: 'sanity.sdk.org-resources',
```
ryanbonial: Could we please use the default tag prefix?

packages/react/src/utils/resolveOrgResources.ts
```diff
@@ -0,0 +1,59 @@
...
+    }).observable,
+  )
+
+  const [mediaLibrariesResult, canvasesResult] = await Promise.allSettled([
+    client.request<OrgResourcesApiResponse>({
+      url: `/media-libraries?organizationId=${encodeURIComponent(organizationId)}`,
```
ryanbonial, line 41:
```suggestion
      url: `/media-libraries?organizationId=${encodeURIComponent(organizationId)}`,
      tag: 'org-resources.media-libraries',
```

packages/react/src/utils/resolveOrgResources.ts
```diff
@@ -0,0 +1,59 @@
...
+    }),
+    client.request<OrgResourcesApiResponse>({
+      url: `/canvases?organizationId=${encodeURIComponent(organizationId)}`,
```
ryanbonial, line 45:
```suggestion
      url: `/canvases?organizationId=${encodeURIComponent(organizationId)}`,
      tag: 'org-resources.canvases',
```

ryanbonial (approved):

PR #791, opened 2026-04-11
fix(core): ensure first event is always present

cngonzalez:
### Description
The Claude review in #643 called out a legitimate bug. The original code:

`const subscription = sharedValues.pipe(skip(1)).subscribe...`

took in a `sharedValues` observable that _should_ have always been composed of a first snapshot event and then subsequent change events: `pipe(map(getCurrent), distinctUntilChanged(isEqual))` 

This would then be shared with `share`. `subscribe`, by design should always skip the initial snapshot.

This could be problematic if there are two subscribers at the same time, and we saw this in the failing tests on that branch:

```ts
// packages/core/src/utils/createFetcherStore.test.ts
const stateSource1 = store.getState(instance, 1)
const sub1 = stateSource1.subscribe()
sub1()


const stateSource2 = store.getState(instance, 1)
const sub2 = stateSource2.subscribe()
// times out, waits forever because it skipped a legitimate state change from its own onSubscribe fetch event
await firstValueFrom(stateSource2.observable.pipe(filter((data) => data !== undefined)))
```

This PR ensures that the first event is always the snapshot (via `shareReplay` with a refCount to ensure we don't keep a stale pipeline), and thus always there for observables and skipped for subscribers.

### What to review
The only actual behavioral change is a one-line change in `createStateSourceAction` which hopefully makes sense.

### Testing
`getProjectionState` tests were updated now that they share a subscription; we didn't really need to keep individual subscriber identities, since we only care if there's _anything_ subscribing to a document / projection.

`getFetcherStore` also has one update for the shared subscription behavior -- the test change actually makes it more consistent with the test directly above it. It was previously tracking the same call to `subscribe` as a "new subscriber" when arguably it is more correct that a new `getState` subscribe call is the new subscriber, triggering the new fetch.

`favorites.test.ts` received a similar update so it's testing the correct situation.

### Fun gif
There are still no fun GIFs about RxJS.

### DIFF
--- a/packages/core/src/favorites/favorites.test.ts
+++ b/packages/core/src/favorites/favorites.test.ts
@@ -146,8 +146,9 @@ describe('favoritesStore', () => {
       await firstValueFrom(state.observable)
       expect(mockFetch).toHaveBeenCalledTimes(1)
       // Second subscriber should use cached response
-      const sub2 = state.subscribe()
-      await firstValueFrom(state.observable)
+      const state2 = getFavoritesState(instance!, mockContext)
+      const sub2 = state2.subscribe()
+      await firstValueFrom(state2.observable)
       expect(mockFetch).toHaveBeenCalledTimes(1)
       // Cleanup
       sub1()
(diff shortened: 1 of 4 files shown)


judofyr (approved):

judofyr:
Aah. That makes a lot of sense! Thanks for really diving into the details here. Merged it into my PR so we can merge that at once.

PR #792, opened 2026-04-15
feat(core): add dev-mode telemetry with consent-gated event tracking

ryanbonial:
### Description

Implements SDK-1133: dev-mode telemetry for the SDK using `@sanity/telemetry`.

NOTE: The React peer dependency in `@sanity/telemetry`'s `package.json` is overly broad. It should be marked as optional since it's only needed if you use the `/react` sub-export. The main entry point works fine without React.

When the SDK is running in a local development environment (`localhost` / `127.0.0.1` + `NODE_ENV !== 'production'`), it emits lightweight adoption events (session start/end, hook first-usage, dev errors) to help understand how the SDK is being used during development.

**Key design decisions:**

- **Consent-gated**: Telemetry only activates after an explicit opt-in check against `GET /intake/telemetry-status`. If the user hasn't opted in, the manager is discarded and zero events are buffered or sent. Unauthenticated users (e.g. public dataset readers) hit a 401 on the consent check which is treated as `undetermined` and also results in no telemetry.
- **Zero production impact**: The telemetry module (`telemetryManager.ts`, `@sanity/telemetry`) is dynamically imported behind the `isDevMode()` gate. Bundlers can tree-shake or code-split the entire path out of production builds.
- **Privacy-safe**: No GROQ queries, document IDs, field names, or error messages are included. Only SDK-internal identifiers (`hookName`, `errorType`) are tracked.
- **Best-effort**: All telemetry operations are wrapped in catch blocks. A failure in telemetry never breaks the SDK.

**New files in `packages/core/src/telemetry/`:**

| File | Purpose |
|---|---|
| `devMode.ts` | `isDevMode()` combining URL + `NODE_ENV` checks |
| `__telemetry__/sdk.telemetry.ts` | Event definitions via `defineEvent` |
| `telemetryManager.ts` | Batched store, consent caching, transport (`POST /intake/batch` + `sendBeacon`) |
| `initTelemetry.ts` | Orchestrator called from `createSanityInstance`, manages per-instance lifecycle |

**New file in `packages/react/src/hooks/helpers/`:**

| File | Purpose |
|---|---|
| `useTrackHookUsage.ts` | Hook + plain function for logging first usage of each hook type per session |

**Modified hooks** (each gets a one-line `trackHookUsage` / `useTrackHookUsage` call):
`useQuery`, `useDocument`, `useDocumentProjection`, `useDocumentPreview`, `useEditDocument`, `usePresence`, `useDocuments`

### What to review

- The consent flow in `initTelemetry.ts` (lines 49-53): does the early-exit on `!consented` feel right, or should we also guard individual `log*` calls?
- Event schema in `sdk.telemetry.ts`: naming should be coordinated with the data-analytics team before merging.
- The `isDevMode()` heuristic in `devMode.ts`: covers `localhost`/`127.0.0.1` but not custom local domains or LAN IPs. Good enough for v1?
- Dynamic import strategy for avoiding circular deps between `initTelemetry` -> `clientStore` -> `createSanityInstance`.

### Testing

Three new test files added:

- `devMode.test.ts` - URL and `NODE_ENV` combinations
- `telemetryManager.test.ts` - consent resolution (granted, denied, unset, network failure), hook deduplication, session lifecycle
- `initTelemetry.test.ts` - integration: dev-mode gating, missing projectId, consent denial skips telemetry, dispose cleanup

### Fun gif
telemetry


joshuaellis (dismissed):
This is great ✨ v clear, can see how to implement it in other projects too 😄

cngonzalez:
How do we feel about merging this into main rather than next? It looks like it's not too dependent on resource architecture and I'd like to keep the v3 PR as light as I can for the "final" merge. I'm handling rebasing next against main already.

ryanbonial:
Rebased to `main` ready for re-review @joshuaellis

packages/core/src/_exports/_internal.ts
```diff
@@ -0,0 +1,14 @@
+export {isStudioConfig} from '../auth/authMode'
```
cngonzalez, line 1: this might be a holdover from next, i'm sorry 🙈 It's nbd but up to you if you want to initialize this in this PR or use the more dedicated commit from `next` for it.

packages/core/src/telemetry/initTelemetry.ts
```diff
@@ -0,0 +1,205 @@
...
+ * imported to avoid circular dependencies and to keep telemetry code
+ * out of production bundles via code splitting.
+ *
+ * The `projectId` must be passed explicitly because the resource
+ * configuration is typically set by the React layer after the
+ * instance has already been created.
+ *
+ * @internal
+ */
+export function initTelemetry(instance: SanityInstance, projectId: string): void {
```
cngonzalez, line 34: do we have concerns about mismatches between the project ID that telemetry initializes with and subsequent hooks that might be targeting other projects?

packages/core/src/telemetry/initTelemetry.ts
```diff
@@ -0,0 +1,205 @@
...
+      if (buffered) {
+        logger.debug('flushing buffered hooks', {hooks: Array.from(buffered)})
+        for (const hookName of buffered) {
+          manager.logHookFirstUsed(hookName)
+        }
+        pendingHooks.delete(instance)
+      }
+
+      const config = instance.config
+      const perspective = typeof config.perspective === 'string' ? config.perspective : 'published'
```
cngonzalez, line 124: Maybe same question as the one I had about projectId at init vs per-hook-use

packages/core/package.json
```diff
@@ -31,7 +31,12 @@
       "import": "./dist/index.js",
       "default": "./dist/index.js"
     },
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./_internal": {
```
cngonzalez, line 35: sorry, same issue as the `_internal` file as a `next` relic

cngonzalez (dismissed):
Generally LGTM! Nothing blocking since it's internal and we can make changes, just some questions

packages/core/src/telemetry/initTelemetry.ts
ryanbonial: The `projectId` is set once when telemetry initializes from the root `ResourceProvider` and stays fixed for the session. All hook events (including hooks inside nested providers targeting different projects) roll up to that same root manager, so they'd be attributed to the root `projectId`.

For v1 I think this is fine since the vast majority of apps use a single project. If we want multi-project telemetry later, the cleanest approach would be a new `SDKResourceMounted` event that fires when each `ResourceProvider` mounts with a distinct `projectId`. That gives us the "are people building multi-project apps?" signal without changing the existing event schema or complicating hook tracking. Tracking as a known limitation for now and I would want to wait for v3's resource work to add the telemetry event.

packages/core/src/_exports/_internal.ts
ryanbonial: We need the `_internal` export for the telemetry code in this PR (`initTelemetry`, `trackHookMounted`), so it makes sense to include it here rather than pull it from next separately.

cngonzalez: Yeah, I'd prefer this be internal too. No worries!

packages/core/src/telemetry/initTelemetry.ts
cngonzalez: sure, makes sense not to implement twice. We'll keep a lookout during / after v3

cngonzalez (approved):

PR #793, opened 2026-04-16
feat(core): respect __SANITY_STAGING__ global variable if present for apiHost

joshuaellis:
### Description

`__SANITY_STAGING__` is a global variable the CLI sets for studios & apps, if it's true we can set the `apiHost` for the client & auth to be `api.sanity.work`, its implemented the same way studio does which means that applications e..g studio/media-library when setting that variable have their SDK code automatically infer the correct host without setting specific settings – one place for everything :)

### What to review

Are the tests sufficient to give us confidence in this change?

### Testing

As a user I should be able to run `SANITY_INTERNAL_ENV=staging sanity dev` and have the SDK infer the correct API hosts

### Fun gif

the-sopranos-columbus-gif

### DIFF
--- a/packages/core/src/auth/authStore.test.ts
+++ b/packages/core/src/auth/authStore.test.ts
@@ -73,6 +73,35 @@ describe('authStore', () => {
       instance?.dispose()
     })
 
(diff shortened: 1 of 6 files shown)


cngonzalez (approved):
looks good to me! also tested locally. thank you! 🙇

PR #794, opened 2026-04-16
fix(core): use mutations endpoint for liveEdit documents

cngonzalez:
### Description
It's my understanding that the Actions API doesn't really work well with `liveEdit` documents -- an explanation in our [internal thread](https://sanity-io.slack.com/archives/C06D0L2HJ94/p1713210390615989) says that it doesn't quite sit well with the draft model. Our current implementation attempts to use the Actions API but ends up creating a draft -- see the below GIF:

Everything looks correct in the optimistic store in the SDK, but the draft is still being created remotely (we just don't see it, as it's not being listened to).

Additionally -- this also errors with the Create action. If you click `Live Edit` in that same interface and click `Create` you'll get the following error:
```
"action index 0: \"attributes._id\" must be prefixed with either \"drafts.\" or \"versions.{bundleId}\""
```
 
This PR updates the document store to use mutations when working with a `liveEdit` document, which is the same strategy the Studio uses.

I also added an initial value affordance there.

### What to review
The most important changes are in the document store itself -- adding a mutations endpoint to the remote change mechanism. `processActions` is also updated not to create `outgoingActions` if it's a `liveEdit` document. (internal actions are fine, since that's the mechanism used to update the optimistic store).

There were some additional guards / tests to avoid batching and dropping things silently.

### Testing
Logic was added to handle mutations in `documentStore.test.ts` and tests were added for this logic. I also added e2e tests to prevent further regressions.

### DIFF
--- a/apps/kitchensink-react/e2e/live-edit-document-editor.spec.ts
+++ b/apps/kitchensink-react/e2e/live-edit-document-editor.spec.ts
@@ -1,3 +1,5 @@
+import {randomUUID} from 'node:crypto'
+
 import {expect, test} from '@repo/e2e'
 
 test.describe('Live edit document editor', () => {
@@ -7,7 +9,7 @@ test.describe('Live edit document editor', () => {
     getPageContext,
   }) => {
     const client = getClient()
-    const liveDocId = `live-e2e-crud-${Date.now()}`
+    const liveDocId = `live-edit-${randomUUID()}`
     // not actually used -- we're checking for non-existence of draft
     const draftId = `drafts.${liveDocId}`
 
--- a/apps/kitchensink-react/e2e/live-edit-document-editor.spec.ts
+++ b/apps/kitchensink-react/e2e/live-edit-document-editor.spec.ts
@@ -1,0 +1,94 @@
+import {expect, test} from '@repo/e2e'
+
+test.describe('Live edit document editor', () => {
+  test('creates, edits, and deletes a published author via editor actions (no draft)', async ({
+    page,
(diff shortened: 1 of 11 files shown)


apps/kitchensink-react/e2e/live-edit-document-editor.spec.ts
```diff
@@ -0,0 +1,94 @@
+import {expect, test} from '@repo/e2e'
+
+test.describe('Live edit document editor', () => {
+  test('creates, edits, and deletes a published author via editor actions (no draft)', async ({
+    page,
+    getClient,
+    getPageContext,
+  }) => {
+    const client = getClient()
+    const liveDocId = `live-e2e-crud-${Date.now()}`
```
ryanbonial: Would this clash on docIDs for our different browser tests that kick off at the same time? Maybe a random UUID instead

ryanbonial:
I think that the stuck state that Claude pointed out is valid. Can you look into a possible fix @cngonzalez?

cngonzalez:
> I think that the stuck state that Claude pointed out is valid. Can you look into a possible fix @cngonzalez?

Done! The current fix opens the possibility that if there's somehow a mixed-batch `liveEdit` and normal set of edits, the latter will go through the mutations API (fine, not actually an error) and also I'm not entirely sure that's possible anyway (Claude, in some of its old comments, seems to think not, but looking at `applyDocumentActions` I believe it might be)

ryanbonial (approved):

PR #802, opened 2026-04-20
fix(document): ensure document store works for versions

cngonzalez:
### Description
This is a cherry-pick from #754, but applied to `main`. There had to be some minor changes now that `liveEdit` documents work a bit differently, but shouldn't be too off. Some additional help text was added for users new to the concepts.

### What to review
Does everything make sense?

### Testing
e2e and unit tests were added.


cngonzalez:
Also if it helps Claude is wrong -- it's useful to have published and draft documents available because they're needed for creation (we base the version off the draft/published) and discard (we replace what you see with the draft / published)

ryanbonial (dismissed):

ryanbonial (approved):

PR #808, opened 2026-04-22
feat(core): make resource name type

cngonzalez:
### Fun gif

FIXES SDK-1091


cngonzalez:
Closing it since not going in this direction.

cngonzalez:
closing -- not really feasible to bring into v2 since we have no way of knowing / serializing what resources folks are using (yet)

PR #809, opened 2026-04-22
feat(core): send _origin, _sdkVersion, _appId on SDK API requests

ryanbonial:
### Description

Implements SDK-1165: the SDK now attaches three query params to every API request it makes through `@sanity/client`:

- `_origin` — `window.location.origin` in the browser, `"server"` otherwise
- `_sdkVersion` — current `@sanity/sdk` version
- `_appId` — the Sanity app ID (when set)

Injection happens inside `@sanity/sdk` via a `get-it` middleware that clones the default `@sanity/client` requester. No changes to `@sanity/client`. Kong's `request-transformer` will strip these before they reach upstream services; a companion Kong change is required for this SDK change to be safe to roll out, since unknown query params cause 400s on services like Mercury.

On `applicationId` specifically: [cli#878](https://github.com/sanity-io/cli/pull/878) and its dependent [cli#961](https://github.com/sanity-io/cli/pull/961) show the CLI using this pattern for `app.resources` ("simpler, with less magic") in favor of a generated `.sanity/resources.ts` module that apps import and pass through. We align with that shape: `SanityConfig.applicationId` is the one wire, and a future CLI release can generate a module that reads `deployment.appId` (or the legacy `app.id`) from `sanity.cli.ts` for apps to import.

Kong PR change: https://github.com/sanity-io/kong/pull/1248

Linear: [SDK-1165](https://linear.app/sanity/issue/SDK-1165)

### What to review

Files of interest:

- `packages/core/src/client/sdkRequester.ts` — new. `createSdkAttributionMiddleware` is the unit of interest; `createSdkRequester` wraps the default requester with it. Types are inferred from `@sanity/client`'s exported `requester` so we don't take a direct dep on `get-it`.
- `packages/core/src/client/clientStore.ts` — resolves `applicationId` from `options ?? instance.config`, includes it in the cache key, strips it from the options handed to `createClient`, and passes the custom requester.
- `packages/core/src/utils/getOrigin.ts` — browser/server origin helper.
- `packages/core/src/config/sanityConfig.ts` — new `applicationId?: string` field on `SanityConfig`. Docstring points at `sanity.cli.ts` as source of truth.
- `apps/kitchensink-react/src/sanityConfigs.ts` — wires `applicationId` through so the network tab shows `_appId`.

A few things worth sanity-checking:

- Middleware ordering: we append after `get-it`'s default `processOptions` so `options.query` has already been encoded into `options.url`. We mutate `options.url` directly.
- Client caching: `applicationId` is in the cache key, so two calls with different app IDs get different `SanityClient` instances (confirmed by test).
- No leak into `@sanity/client` config: `applicationId` is stripped from `createClient` options before being passed. Confirmed by test.

### Testing

- `packages/core/src/client/sdkRequester.test.ts` — covers middleware behavior: appends params, preserves existing query strings, omits `_appId` when unset, handles both browser and server origins.
- `packages/core/src/utils/getOrigin.test.ts` — browser/server branches.
- `packages/core/src/client/clientStore.test.ts` — updates cover the `applicationId` cache-key behavior, distinct client instances per app ID, and that `applicationId` does not leak into `createClient`'s config.

Local verification:

- Manual: ran the kitchensink dev server and confirmed requests carry `_origin`, `_sdkVersion`, and `_appId=wkyoigmzawwnnwx458zgoh46` in the query string.

### Notes for release

**Blocked on a Kong change.** The SDK now sends these params on every request, including to endpoints that don't currently tolerate unknown query params (e.g. `GET /v*/projects/*` on Mercury returns 400). The Kong `request-transformer` plugin must be updated to strip `_origin`, `_sdkVersion`, and `_appId` from the query string before proxying upstream. That change should land (and be verified in staging) before this SDK release is cut.

### Fun gif
be prepared

### DIFF
--- a/apps/kitchensink-react/src/sanityConfigs.ts
+++ b/apps/kitchensink-react/src/sanityConfigs.ts
@@ -1,13 +1,20 @@
 import {type DocumentSource, type SanityConfig} from '@sanity/sdk'
 
+// Mirrors `app.id` in sanity.cli.ts. A future CLI release is expected to emit a
+// generated module (alongside `.sanity/resources.ts`) that we can import here
+// instead of hardcoding.
+const APPLICATION_ID = 'wkyoigmzawwnnwx458zgoh46'
+
 export const devConfigs: SanityConfig[] = [
   {
     projectId: 'ppsg7ml5',
     dataset: 'test',
+    applicationId: APPLICATION_ID,
   },
   {
     projectId: 'vo1ysemo',
     dataset: 'production',
+    applicationId: APPLICATION_ID,
   },
 ]
 
--- a/packages/core/src/client/clientStore.test.ts
+++ b/packages/core/src/client/clientStore.test.ts
@@ -7,7 +7,13 @@ import {createSanityInstance, type SanityInstance} from '../store/createSanityIn
 import {getClient, getClientState} from './clientStore'
 
 // Mock dependencies
-vi.mock('@sanity/client')
+vi.mock('@sanity/client', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@sanity/client')>()
+  return {
+    ...actual,
+    createClient: vi.fn(),
+  }
+})
 
 vi.mock('../auth/authStore')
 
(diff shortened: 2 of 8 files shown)


ryanbonial:
Closing due to the issues with cace invalidation for content lake that these params woudl cause

PR #810, opened 2026-04-22
fix: keep LoginError synchronous in standalone apps

ryanbonial:
### Description

Fixes [SDK-1318](https://linear.app/sanity/issue/SDK-1318). In standalone SDK apps (using `SDKProvider` outside the Sanity Dashboard), the app got stuck on `SDKProvider`'s Suspense fallback whenever an auth error occurred, instead of surfacing the error or redirecting to `www.sanity.io/login`.

Root cause: `LoginError` unconditionally called `useWindowConnection` at the top level. That hook suspends until a Comlink node to the parent window is established. In standalone mode there is no parent window, so the suspension never resolved and bubbled up to `SDKProvider`'s outer `<Suspense>` boundary. The underlying auth store was transitioning to `ERROR` correctly (visible in Redux DevTools), but the UI never re-rendered to react to it.

Changes:

- New `DashboardAccessRequest` component that owns the only caller of `useWindowConnection` inside `LoginError` (the `dashboard/v1/auth/access/request` post to the parent window for projectUserNotFound errors).
- `LoginError` now renders synchronously in all environments. It checks `getIsInDashboardState(instance).getCurrent()` and only mounts `DashboardAccessRequest` (inside a local `<Suspense fallback={null}>`) when actually running in the dashboard. In standalone apps nothing suspends, so `AuthBoundary`'s redirect and error UI can do their job.

### What to review

- `packages/react/src/components/auth/LoginError.tsx`: verify the gating logic (`isProjectUserNotFound && projectId && getIsInDashboardState(instance).getCurrent()`) covers the intended dashboard flow, and that the local `<Suspense fallback={null}>` is the right boundary.
- `packages/react/src/components/auth/DashboardAccessRequest.tsx`: new internal component. The doc comment explains why it must be rendered under its own Suspense boundary.
- Affected flows: the auth error surface in the Dashboard (project-access-request handoff) and the equivalent surface in standalone apps (should render the error + retry CTA or let `AuthBoundary` redirect to login).

### Testing

Unit tests added in `LoginError.test.tsx`:

- `renders synchronously on a 401 projectUserNotFound error outside the dashboard`: asserts that the Suspense fallback is never shown and `useWindowConnection` is not called.
- `fires the dashboard access request on a 401 projectUserNotFound error inside the dashboard`: asserts that `fetch('dashboard/v1/auth/access/request', ...)` still runs with the expected payload when in-dashboard.
- Existing `AuthError`/rethrow tests still pass.

Manual repro used a minimal standalone app (`SDKProvider` + `useCurrentUser`) with a stamped-looking but unauthorized token seeded into `localStorage.__sanity_auth_token`. Before the fix: app stayed on the `SDKProvider` Suspense fallback. After the fix: `LoginError` renders, the 401 triggers `handleRetry` → `logout()` → `LOGGED_OUT`, and `AuthBoundary` redirects to `www.sanity.io/login` as expected.

### Fun gif
suspenders

### DIFF
--- a/packages/react/src/components/auth/LoginError.tsx
+++ b/packages/react/src/components/auth/LoginError.tsx
@@ -57,9 +57,12 @@ export function LoginError({error, resetErrorBoundary}: LoginErrorProps): React.
   // The dashboard access request flow relies on a comlink connection to the
   // parent window. In standalone apps that connection never materializes, so
   // we must skip it entirely to avoid suspending forever on the parent's
-  // Suspense boundary (see SDK-1318).
-  const shouldRequestDashboardAccess =
-    isProjectUserNotFound && !!projectId && getIsInDashboardState(instance).getCurrent()
+  // Suspense boundary (see SDK-1318). Resolving to the projectId (or null)
+  // here lets the JSX render the child with a single non-null guard.
+  const dashboardAccessProjectId =
+    isProjectUserNotFound && projectId && getIsInDashboardState(instance).getCurrent()
+      ? projectId
+      : null
 
   const handleRetry = useCallback(async () => {
     await logout()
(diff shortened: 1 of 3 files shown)


packages/react/src/components/auth/LoginError.tsx
```diff
@@ -94,20 +94,29 @@ export function LoginError({error, resetErrorBoundary}: LoginErrorProps): React.
...
+        description={authErrorMessage}
+        cta={
+          showRetryCta
+            ? {
+                text: 'Retry',
+                onClick: handleRetry,
+              }
+            : undefined
+        }
+      />
```
gu-stav, line 119: With the usage of the `SDKProvider` is there a way to go straight to the login, without surfacing the error message? I've tampered with my token to be invalid, reloaded and now I see an error, but it would be better UX-wise I think if I would land on the login page again.

gu-stav (approved):
The suspense boundary properly resolves now and surfaces the error. ✅

packages/react/src/components/auth/LoginError.tsx
ryanbonial: You're right this sin't handled quite correctly. I'll make a followup PR to clean up the Error message display.

PR #814, opened 2026-04-24
fix(react): guard LoginError auto-logout against re-entry

ryanbonial:
### Description

Redirects standalone SDK apps to `www.sanity.io/login` when the stored token returns a non-`projectUserNotFound` 401, instead of leaving the user parked on a generic "Authentication Error" screen. This is the follow-up Gustav asked for on [#810](https://github.com/sanity-io/sdk/pull/810#discussion_r3129027028) and closes SDK-1323.

Two things changed in `LoginError`:

1. The `ClientError` is now unwrapped from `AuthError.cause`. `AuthBoundary` always wraps the store's `ClientError` in an `AuthError` before the error boundary sees it, so the existing `error instanceof ClientError` checks (for 401 `projectUserNotFound`, 404 session errors, and the dashboard access request path added in #810) were dead in production. Unwrapping makes them reachable.
2. For a non-`projectUserNotFound` 401 in a standalone app, `LoginError` now briefly renders "Signing you out and returning to login..." and kicks off `logout()`. That flips the store to `LOGGED_OUT`, which `AuthBoundary`'s existing effect turns into a hard redirect to the Sanity login page. The auto-logout is guarded by a `useRef` one-shot so the effect can't re-enter itself when `useAuthState` emits the new `LOGGED_OUT` reference (without the guard React hits "Maximum update depth exceeded" and blanks the page).

Dashboard-embedded apps keep their existing behavior:

- 401 generic: UI stays on the error screen with a Retry button; `ComlinkTokenRefreshProvider` continues to request a new token from the parent window off the `ERROR`+401 state on its own.
- 401 `projectUserNotFound`: the `DashboardAccessRequest` comlink call now actually fires (this was the broken path from #810 / SDK-1318).
- 404 session invalid: descriptive "session ID is invalid" / "login link is invalid" message now shows in the dashboard too, rather than the generic fallback.
- `ConfigurationError` (org mismatch): unchanged.

### What to review

- `packages/react/src/components/auth/LoginError.tsx`: the `AuthError.cause` unwrap, the `isInDashboard` gate, and the `hasAutoLoggedOutRef` one-shot around the standalone `handleRetry()` call.
- Dashboard-side smoke test: open any dashboard-embedded app, force a 401 (expired/revoked token, or a user who isn't a member of the project). Expect the existing token-refresh behavior to keep working and the `projectUserNotFound` flow to now invoke the dashboard access request.
- Standalone-side smoke test with `apps/standalone-react`: click **Expire token**, confirm you see the "Signing you out..." message briefly and are redirected to `www.sanity.io/login`, not a blank page.

### Testing

- `packages/react/src/components/auth/LoginError.test.tsx` gets three new cases: standalone auto-logout on a generic 401, dashboard non-auto-logout on a generic 401, and the dashboard `projectUserNotFound` path driven through the real `AuthError(ClientError)` wrapping used in production.
- The existing `AuthBoundary` and `ComlinkTokenRefresh` test suites still pass unchanged, covering the `LOGGED_OUT` redirect, the iframe gate, and the dashboard token-refresh flow this PR relies on.
### Fun gif
try again


gu-stav (approved):
Nice, thank you!

PR #815, opened 2026-04-24
chore: Fallow for monorepo code quality

ryanbonial:
### Description

Adopt Fallow as a static-analysis gate alongside knip. Fallow covers what knip and ESLint don't: cross-package duplication, cyclomatic and cognitive complexity, churn-weighted hotspots, ranked refactor targets, and priority-weighted cycle detection. Knip stays in place as the dead-code and unused-dependency gate.

The CI workflow blocks on _new_ findings only. Three baseline files (`fallow-baselines/{dead-code,health,dupes}.json`) suppress everything that already exists on `main`, so day-one PRs see "audit passed" unless they actually regress something. The baselines were regenerated against current `main` after the latest merge. Knip, ESLint, Prettier, TypeScript, the existing `pnpm depcheck` script, and the `Depcheck` CI workflow are unchanged.

What's in the branch:

* `fallow@^2.84.0` pinned as a devDependency. 2.84 added a built-in pkg-utils plugin, so the config no longer needs the entry-point and ignore-pattern boilerplate from earlier revisions (thanks @BartWaardenburg)
* Two pnpm scripts (down from nine, per review feedback): `fallow` as a passthrough for ad-hoc use, and `fallow:baseline` to regenerate all three baselines. There is no single fallow command that saves all baselines, so that script chains the three `--save-baseline` invocations
* `.fallowrc.json`: one ignore pattern (`**/typedoc-theme/**`) plus the `audit` block pointing at the three baseline files
* `.github/workflows/fallow.yml`: blocking job that audits changed files against the PR base, propagates Fallow's exit code, and renders the audit, summary, and hotspot output into the job's step summary. Uses the same SHA-pinned `actions/checkout@v6` as the other workflows. Summary and hotspots call the `fallow` passthrough directly
* `fallow-baselines/` with three baseline JSONs and a README explaining what each file suppresses, when to regenerate, and when not to
* `.fallow/` added to `.gitignore` for Fallow's local cache

What this enables, with concrete examples already in this repo:

* One real cycle in `packages/core` (`queryStore` → `getPerspectiveState` → `releasesStore`) currently suppressed by `eslint-disable import/no-cycle` comments. Fallow ranks it as the top refactor target by fan-in × churn
* 278 baselined clone groups (10.8% duplication), including a byte-identical `getEnv.ts` between `core` and `react` and a duplicated RxJS subscription helper shared by `queryStore` and `usersStore`
* 55 functions over the complexity thresholds and 7 ranked refactor targets, e.g. `insert` in `patchOperations.ts` at cognitive complexity 124
* The CI workflow's step summary surfaces all of the above on every PR

### What to review

1. `.github/workflows/fallow.yml` — the gate. Audit is blocking; Hotspots reports under `if: always()` so reviewers still see context if the audit fails
2. `.fallowrc.json` — the config, including the `audit` block that wires baselines to the workflow
3. `fallow-baselines/README.md` — the operational guide for the team. Calls out the dupes line-anchor caveat, when to regenerate baselines, and when _not_ to (don't regenerate to silence a real regression)
4. `fallow-baselines/dead-code.json`, `health.json`, `dupes.json` — what's currently masked. Skim to confirm nothing surprising is in the baseline; entries here are tech debt the team has implicitly accepted as of this commit

Locally:

    pnpm install
    pnpm fallow                          # full run: dead-code + dupes + health
    pnpm fallow audit --base origin/main # what CI runs; exits 0 on this branch
    pnpm fallow health --hotspots        # churn × complexity
    pnpm fallow health --targets         # ranked refactor backlog
    pnpm fallow:baseline                 # regenerate all three baselines after fixing existing findings

Notes for after this lands:

* First PRs against `main` will see "audit passed" unless they introduce new findings. That's the intended UX
* The dupes baseline is line-anchored. Editing above a baselined clone shifts the entry's anchor and the clone re-surfaces as "new" duplication. The README documents this; expect occasional `pnpm fallow:baseline` regenerations as part of normal merge-to-main flow

### Testing

No runtime code changes, so no automated tests. Validation:

* `pnpm fallow audit --base origin/main` exits 0 on this branch with all three baselines wired
* The audit step in CI propagates Fallow's exit code (no `|| true` swallowing failures); a synthetic regression on a scratch branch (duplicated `getEnv`, deeply nested function, unreachable file) was caught end-to-end with non-zero exit
* `pnpm fallow:baseline` regenerates all three files; reruns produce semantically identical output (only prettier formatting differs from the committed versions)
* Knip output (`pnpm depcheck`) verified unchanged on this branch
 
### Fun gif

fallow kids


cngonzalez:
I'm a little nervous about the package -- it seems to be only a month old (judging from Github, npm is down right now), has a postinstall script for a binary, but I don't actually see the binaries in its repo.

Is this urgent? Happy to make tickets for what it's found (though it's stuff I've noted explicitly, happy to focus on those issues if  they're problematic) but I personally would be more comfortable waiting until this is a bit more established? We've run into issues with the past with newer / less-popular packages.

Also happy to be wrong, I don't mean to block anything.

ryanbonial:
The postinstall is verifies that he right platform of the binary is used, the (rust) binary itself is pulled in during the install. However, your points still stand it is a pretty new tool so I'm ok with letting this PR bake for a while.

cngonzalez:
Appreciated! I'm going to unassign myself so I can keep my PR sidebar clear but feel free to reassign me a bit further on

BartWaardenburg:
Thanks for the thorough adoption PR @ryanbonial. The `package.config.ts` / `package.bundle.ts` entries are exactly the sort of `@sanity/pkg-utils` paper-cut that shouldn't need hand-config, so v2.84.0 adds a built-in `pkg-utils` plugin that handles them. Together with a few longstanding defaults, your `.fallowrc.json` can shrink a lot.

Bump to `fallow@^2.84.0`, then:

**The whole `entry` array can go:**
- `package.config.ts`, `packages/*/package.config.ts`, and `packages/*/package.bundle.ts` are now kept reachable by the `pkg-utils` plugin. It activates on your `@sanity/pkg-utils` dependency and covers the repo root plus every workspace package.
- `scripts/*.{ts,mts}` are already reachable without config: fallow reads the `tsx scripts/*.mts` invocations in your root `package.json` scripts and treats those files as entry points.

**Most of `ignorePatterns` duplicates built-in defaults:**
- `**/dist/**` and `**/coverage/**` are default-ignored.
- `**/build/**`: root `build/` is a default, and there are no nested `build/` dirs in source.
- `**/.turbo/**`: dotfile directories outside fallow's allowlist are already skipped.
- `**/lib/**` and `**/docs/**` match nothing in the source tree (pkg-utils builds to `dist/`).
- Keep `**/typedoc-theme/**`. That one is doing real work (`typedoc-theme/style.css` is otherwise flagged).

**`workspaces.apps/kitchensink-react.plugins.playwright: false`** is a no-op on 2.84.0; removing it produces identical findings, so the whole `workspaces` block can go too.

That leaves:
```json
{
  "$schema": "https://raw.githubusercontent.com/fallow-rs/fallow/main/schema.json",
  "ignorePatterns": ["**/typedoc-theme/**"],
  "audit": {
    "deadCodeBaseline": "fallow-baselines/dead-code.json",
    "healthBaseline": "fallow-baselines/health.json",
    "dupesBaseline": "fallow-baselines/dupes.json"
  }
}
```

I ran your branch with both the current and the trimmed config and the findings are identical, so this is a safe cleanup rather than a behavior change. Thanks again for the detailed writeup in the PR description, it is genuinely useful signal.

package.json
```diff
@@ -33,6 +33,15 @@
     "dev:e2e": "SANITY_INTERNAL_ENV=staging SDK_E2E_ORGANIZATION_ID=oFvj4MZWQ turbo run dev --filter=@sanity/kitchensink-react",
     "dev:kitchensink": "turbo run dev --filter=@sanity/kitchensink-react",
     "dev:standalone": "turbo run dev --filter=@sanity/standalone-react",
+    "fallow": "fallow",
```
cngonzalez, line 36: should we limit the amount of scripts here to only those used in CI? this is a bit overwhelming and somewhat dwarfs any other script area

cngonzalez (changes requested):
a nit about scripts, let me know what you think

package.json
cngonzalez: Ah, I suppose we can't because it's all top-level. I wonder if a `@repo/fallow` package might be helpful to keep things a bit tidier, as we've done for ailf?

ryanbonial: These can definitely be pruned down 👍

cngonzalez: thank you! 💜

cngonzalez (approved):

PR #819, opened 2026-04-28
feat: handle chunk-load errors with automatic reload

ryanbonial:
### Description

Resolves SDK-1345.

Users on a stale tab after a deploy hit chunk-load errors when their browser tries to fetch a hashed JS or CSS chunk that no longer exists at the CDN. Studio already detects these via an `isImportError` helper and tells the user to reload; the SDK had no equivalent and the page just broke.

This adds the same kind of handling, in an SDK-shaped way:

- `isImportError` in `@sanity/sdk` (`@public`) detects the common shapes: webpack `ChunkLoadError`, Vite `Failed to fetch dynamically imported module`, Firefox `error loading dynamically imported module`, Safari `Importing a module script failed`, and Vite `Unable to preload CSS`.
- A new top-level `ErrorBoundary` inside `SDKProvider` catches those errors above every `ResourceProvider` and `AuthBoundary`. On the first occurrence in a session it triggers `window.location.reload()` and renders nothing (so the user doesn't see a flash of error UI before the navigation). A `sessionStorage` flag (`__sanity_sdk_chunk_reload_attempted`, matching the existing `__sanity_auth_token` underscore convention) prevents reload loops; if a reload has already been attempted in the same session the SDK falls back to a manual "Reload page" UI built from the existing internal `Error` component.
- `AuthBoundary`'s fallback gets a defensive `isImportError` short-circuit so chunk errors thrown beneath it (typically lazy-loaded code in the consumer's app) get the same fallback instead of being misreported as auth errors.
- The flag is cleared once the SDK successfully mounts past the boundary, so a future incident in the same session can trigger another automatic reload.

API surface stays minimal: no override prop on `SanityApp` (matches how `CorsErrorComponent` is wired today), and the `ChunkLoadError` component is `@internal`. Only `isImportError` is part of the public API for consumers who want to apply the same detection inside their own route-level error boundaries.

### What to review

- `[packages/core/src/utils/isImportError.ts](packages/core/src/utils/isImportError.ts)`: the regex set is the main correctness call. Verified against current Vite docs (vitejs/vite#20313, merged Aug 2025), webpack/webpack#7502 / #16618, and the WebKit Safari bug (#242740) referenced in withastro/astro#14775 and sveltejs/kit#5208. Worth a sanity check on the message strings and the case-insensitive regex flag choices.
- `[packages/react/src/components/SDKProvider.tsx](packages/react/src/components/SDKProvider.tsx)`: the new `ErrorBoundary` wraps the existing nested-providers tree. `ChunkAwareFallback` rethrows non-import errors so `AuthBoundary`'s existing boundary keeps handling auth/CORS. `ResetChunkReloadFlagOnMount` clears the flag on successful mount.
- `[packages/react/src/components/auth/AuthBoundary.tsx](packages/react/src/components/auth/AuthBoundary.tsx)`: defensive `isImportError` branch added at the top of the existing fallback. Safe duplication; `AuthBoundary` is the closest boundary to the consumer's children, so it catches first.
- `[packages/react/src/components/errors/ChunkLoadError.tsx](packages/react/src/components/errors/ChunkLoadError.tsx)` and `[chunkReloadStorage.ts](packages/react/src/components/errors/chunkReloadStorage.ts)`: storage helpers split into a sibling file so the component file only exports React components (keeps Fast Refresh happy). All storage access is `try`/`catch` plus `typeof window` guards for SSR / private-mode quotas.

Affected surface area: any consumer of `@sanity/sdk-react` that uses `<SanityApp>` or `<SDKProvider>`. No prop changes, no new dependencies (`react-error-boundary` was already there).

Two reviewer-facing decisions worth confirming:

1. **Auto-reload-once vs. manual UI as the default.** Picked auto-reload-once because chunk-load errors are almost always stale-tab-after-deploy (a fresh `index.html` fixes it), and silently fixing the user is the dominant industry pattern. Loop guard means the manual UI still appears if reload doesn't help.
2. **Two boundaries (SDKProvider + AuthBoundary).** Kept both. `AuthBoundary` catches the common case (errors from consumer code); `SDKProvider`'s boundary is a backstop for chunk errors thrown above `AuthBoundary` (e.g. from inside a `ResourceProvider`). Negligible runtime cost.

### Testing

- New `[isImportError.test.ts](packages/core/src/utils/isImportError.test.ts)` covers each detected error shape (webpack name + numeric and named chunks, Vite, Firefox variant, Safari with and without the `ing` suffix, Vite CSS preload) plus negative cases (non-Error values, unrelated errors) and case insensitivity. 10 tests, all green.
- New `[ChunkLoadError.test.tsx](packages/react/src/components/errors/ChunkLoadError.test.tsx)` mocks `window.sessionStorage` and stubs `window.location.reload` to assert (a) first occurrence triggers reload and sets the flag, rendering nothing; (b) second occurrence with the flag set renders the manual UI, and clicking "Reload page" both clears the flag and reloads.
- Existing `SDKProvider`, `SanityApp`, and `AuthBoundary` tests still pass without modification, confirming the new boundary doesn't break the existing render contract.
- Full workspace `pnpm test:coverage` run: 1123 passed, 3 skipped, no regressions. New files coverage: `isImportError.ts` 100/100/100/90, `ChunkLoadError.tsx` 96.96/100/96.96/85.71, `chunkReloadStorage.ts` 76/100/76/33.33 (uncovered branches are the SSR / no-`sessionStorage` / storage-throws guards; happy to add stubs for those if you'd like the helper module up to threshold). Workspace totals stay above the configured 95/95/95/90.
- Manual sanity check left to the reviewer: throw `new Error('Failed to fetch dynamically imported module')` from a component beneath `<SanityApp>` and confirm the page reloads once, then renders the fallback UI on the second throw.

### Fun gif

chunk


cngonzalez (approved):
LGTM, thanks for taking it on!

PR #826, opened 2026-04-30
feat(react): add SanityInstanceProvider for externally-created instances

joshuaellis:
## Summary

- Adds `SanityInstanceProvider` — a new public component that accepts a pre-created `SanityInstance` and provides it to the React tree via context + Suspense
- Refactors `ResourceProvider` to use `SanityInstanceProvider` internally (no behavior change)

## Use case

`SanityApp` and `ResourceProvider` always create their own `SanityInstance` internally. There's no way to provide an instance created outside the React tree.

This forces an architectural inversion for apps where a non-React system layer owns the SDK instance — you end up with duplicated state, duplicated auth, and no single source of truth.

`SanityInstanceProvider` sits at the bottom of the provider ladder (`SanityApp → SDKProvider → ResourceProvider → SanityInstanceProvider`) and gives full control to the caller:

```tsx
const instance = createSanityInstance({ projectId: 'abc', dataset: 'production' })

<SanityInstanceProvider instance={instance} fallback={<Loading />}>
  <App /> {/* useSanityInstance(), useDocuments(), etc. all work */}
</SanityInstanceProvider>
```

The caller owns the instance lifecycle — `SanityInstanceProvider` will not dispose it on unmount.

## Open questions

- Should `initTelemetry` live in `SanityInstanceProvider` rather than `ResourceProvider`? Currently telemetry is only initialised when `ResourceProvider` creates a root instance. If someone uses `SanityInstanceProvider` directly, telemetry won't be initialised automatically. This might be the right default (caller controls everything), but worth discussing. The current implementation also specifically won't happen if theres a parent sanity instance – so we probably do want to keep it in ResourceProvider?

### DIFF
--- a/packages/react/src/context/SanityInstanceProvider.tsx
+++ b/packages/react/src/context/SanityInstanceProvider.tsx
@@ -25,7 +25,7 @@ export interface SanityInstanceProviderProps {
 /**
  * Provides an externally-created Sanity instance to child components through React Context.
  *
- * @public
+ * @internal
  *
  * @remarks
  * Unlike {@link ResourceProvider}, this component does not create or dispose a SanityInstance.
--- a/packages/react/src/context/SanityInstanceProvider.tsx
+++ b/packages/react/src/context/SanityInstanceProvider.tsx
@@ -1,0 +1,71 @@
+import {type SanityInstance} from '@sanity/sdk'
+import {Suspense} from 'react'
+
+import {SanityInstanceContext} from './SanityInstanceContext'
+
+/**
+ * Props for the SanityInstanceProvider component
+ * @public
+ */
+export interface SanityInstanceProviderProps {
+  /**
+   * A pre-created SanityInstance to provide to child components.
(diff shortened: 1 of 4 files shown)


cngonzalez (dismissed):
looks reasonable to me!

packages/react/src/context/SanityInstanceProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * Used as the fallback for the internal Suspense boundary.
+   */
+  fallback: React.ReactNode
+  children: React.ReactNode
+}
+
+/**
+ * Provides an externally-created Sanity instance to child components through React Context.
+ *
+ * @public
```
ryanbonial: Would this likely only be used for other Sanity products? if so, marking it as internal would be good because using it could result in odd behavior.

```suggestion
 * @internal
```

joshuaellis: I think we can mark as internal for now. Good idea.

cngonzalez (approved):

PR #834, opened 2026-05-01
refactor: turn project / dataset pairs into resources internally

cngonzalez:
### Description
This PR takes some of the logic from the old v3 branch in a non-breaking way. In anticipation of not having an instance inheritance system, or the instance as the source of truth for configuration, this PR boosts the resource normalization logic to use resources in context and only rely on the instance configuration as a final step. This will make moving to v4 easier since most instance logic now lives in only a few files.

This doesn't change functionality -- all actions are bound by resources (or otherwise not bound to instance config) so there's no need for finding specific instances.

### What to review
I've noted those places where I made choices that I could use feedback on, but really this is just rearranging parameters. Let me know if there's places I missed or where my logic is wrong.

### Testing
Tests were updated, added, or removed as needed.

### DIFF
--- a/packages/core/src/client/clientStore.test.ts
+++ b/packages/core/src/client/clientStore.test.ts
@@ -191,29 +191,7 @@ describe('clientStore', () => {
   })
 
   describe('resource handling', () => {
-    it('should create client when resource is provided', () => {
-      const client = getClient(instance, {
-        apiVersion: '2024-11-12',
-        resource: {projectId: 'source-project', dataset: 'source-dataset'},
-      })
-
-      expect(vi.mocked(createClient)).toHaveBeenCalledWith(
-        expect.objectContaining({
-          apiVersion: '2024-11-12',
-          resource: {type: 'dataset', id: 'source-project.source-dataset'},
-        }),
-      )
-      // Client should be projectless - no projectId/dataset in config
-      expect(client.config()).not.toHaveProperty('projectId')
-      expect(client.config()).not.toHaveProperty('dataset')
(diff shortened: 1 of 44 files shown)


packages/core/src/client/clientStore.test.ts
```diff
@@ -257,38 +235,24 @@ describe('clientStore', () => {
       )
     })
 
-    it('should create projectless client when resource is provided, ignoring instance config', () => {
+    it('should transform dataset resource to project-based config for now', () => {
```
cngonzalez, line 238: With having:
1. projectId / dataset pairs AS WELL as resources in the public API
2. CORS and Studio cookie issues

it doesn't make sense to throw this anymore -- we should try to put people on the happy path, IMO. I don't think there's a compelling reason for a user to need to use the global route for datasets. we can try to rectify this but it'd be better to do after disallowing projectId / dataset parameters (in favor of resources), IMO.

packages/core/src/document/documentStore.ts
```diff
@@ -396,8 +396,7 @@ const subscribeToAppliedAndSubmitNextTransaction = ({
       withLatestFrom(
         getClientState(instance, {
           apiVersion: API_VERSION,
-          // TODO: remove in v3 when we're ready for everything to be queried via resource
-          resource: resource && !isDatasetResource(resource) ? resource : undefined,
+          resource,
```
cngonzalez, line 399: handled by the client store, no need to dupe logic

joshuaellis (approved):

PR #850, opened 2026-05-04
chore(ci): use org level linear api key

binoy14:
### Description

Use the org level linear key instead of a PAT

### What to review

Changes makes sense, verified the token exits on org level but might have to add permissions to it, but hopefully it just works

### Testing

Tested in a release


binoy14:

- #850 ◀

joshuaellis (dismissed):

joshuaellis (approved):

PR #865, opened 2026-05-06
feat(core): add other param options to projectsStore

joshuaellis:
### Description

- Adds `includeFeatures`, `onlyExplicitMembership`, and `organizationId` as typed options to the projects store, replacing the previous minimal inline types with a full `ProjectsOptions` generic interface
- Introduces conditional `Project<IncludeMembers, IncludeFeatures>` type so `members` and `features` fields are present/absent based on the query flags
- Switches from `client.projects.list()` to a raw `client.request()` against `/projects` to pass all query params directly
- Adds type-level tests (`.test-d.ts`) for both `@sanity/sdk` and `@sanity/sdk-react` to verify generic narrowing across all flag combinations
- Exports new public types: `Project`, `ProjectBase`, `ProjectMember`, `ProjectMemberRole`, `ProjectMetadata`, `ProjectsOptions`

### What to review

* Do we see any breaking changes here? I've preserved the current behaviour with the correct defaults and depreacted the type we used to export from the react layer.
* Be interested if we want to start branding types e.g. `id` instead of `string` would be `ProjectId` – we could then in theory developer a better DX for hooks like `access` etc. by narrowing the resource type to a specific form of `id`.

### Fun gif

looking-around-adam-scott-gif-by-apple-tv

### DIFF
--- a/packages/core/src/project/project.ts
+++ b/packages/core/src/project/project.ts
@@ -2,40 +2,161 @@ import {switchMap} from 'rxjs'
 
 import {getClientState} from '../client/clientStore'
 import {type ProjectHandle} from '../config/sanityConfig'
+import {type SanityInstance} from '../store/createSanityInstance'
+import {type StateSource} from '../store/createStateSourceAction'
 import {createFetcherStore} from '../utils/createFetcherStore'
 
 const API_VERSION = 'v2025-02-19'
 
+/** @public */
+export interface ProjectMemberRole {
+  name: string
+  title: string
+  description: string
+}
+
+/** @public */
+export interface ProjectMember {
+  id: string
+  createdAt: string
+  updatedAt: string
+  isCurrentUser: boolean
+  isRobot: boolean
+  roles: ProjectMemberRole[]
+}
+
+/** @public */
+export interface ProjectMetadata {
+  color?: string
+  externalStudioHost?: string
+  initialTemplate?: string
+  cliInitializedAt?: string
+  integration: 'manage' | 'cli'
+}
+
+/**
+ * The base fields returned from `/projects` for every project.
+ * @public
+ */
+export interface ProjectBase {
+  id: string
+  displayName: string
+  studioHost: string | null
+  organizationId: string
+  metadata: ProjectMetadata
+  isBlocked: boolean
+  isDisabled: boolean
+  isDisabledByUser: boolean
+  activityFeedEnabled: boolean
+  createdAt: string
+  updatedAt: string
+}
+
+/**
+ * A `Project` with `members` and/or `features` conditionally included
+ * based on the query options used to fetch it.
+ * @public
+ */
+export type Project<
+  IncludeMembers extends boolean = true,
+  IncludeFeatures extends boolean = true,
+> = ProjectBase &
+  // `boolean extends T` is non-distributive — true only when T is the wide
+  // `boolean`, in which case the field is optional. Literal `true`/`false`
+  // fall through to the strict branch.
+  (boolean extends IncludeMembers
+    ? {members?: ProjectMember[]}
+    : IncludeMembers extends true
+      ? {members: ProjectMember[]}
+      : unknown) &
+  (boolean extends IncludeFeatures
+    ? {features?: string[]}
+    : IncludeFeatures extends true
+      ? {features: string[]}
+      : unknown)
+
+/** @public */
+export interface ProjectOptions<
+  IncludeMembers extends boolean = true,
+  IncludeFeatures extends boolean = true,
+> extends ProjectHandle {
+  includeMembers?: IncludeMembers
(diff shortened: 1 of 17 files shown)


packages/core/src/project/project.ts
```diff
@@ -2,40 +2,161 @@ import {switchMap} from 'rxjs'
...
+    const featuresKey = includeFeatures ? ':features' : ''
+    return `project:${projectId}${membersKey}${featuresKey}`
+  },
+  fetcher: (instance) => (options?: ProjectOptions<boolean, boolean>) => {
+    const projectId = resolveProjectId(instance, options)
+
+    return getClientState(instance, {
+      apiVersion: API_VERSION,
+      scope: 'global',
+      projectId,
```
cngonzalez: is there any reason to pass the project here if we're not using the project hostname? I think it gets cleaned out of the options in `clientStore` with `scope: 'global'` anyway.

packages/core/src/projects/projects.test.ts
```diff
@@ -36,41 +34,119 @@ describe('projects', () => {
...
+  const mockGetKey = (
+    _instance: SanityInstance,
+    options?: {
+      organizationId?: string
+      includeMembers?: boolean
+      includeFeatures?: boolean
+      onlyExplicitMembership?: boolean
+    },
+  ) => {
+    const organizationId = options?.organizationId
```
cngonzalez: should we just declare the key creation function independently in `projects.ts` (i.e., something like `projectsFetcherCacheKey`) and import it here and also set it as the `getKey` function? might be easier to maintain and also would feel like we're "actually" testing it

packages/core/src/projects/projects.test.ts
```diff
@@ -36,41 +34,119 @@ describe('projects', () => {
...
+    const includeFeatures = options?.includeFeatures ?? true
+    const onlyExplicitMembership = options?.onlyExplicitMembership ?? false
+    const orgKey = organizationId ? `:org:${organizationId}` : ''
+    const membersKey = includeMembers ? ':members' : ''
+    const featuresKey = includeFeatures ? ':features' : ''
+    const explicitKey = onlyExplicitMembership ? ':explicit' : ''
+    return `projects${orgKey}${membersKey}${featuresKey}${explicitKey}`
+  }
+
+  const mockInstance = {} as SanityInstance
```
cngonzalez: i always prefer a "real" instance but it's nbd

packages/react/src/hooks/projects/useProjects.ts
```diff
@@ -39,18 +31,18 @@ type UseProjects = <TIncludeMembers extends boolean = false>(options?: {
  * ```
  * @example
  * ```tsx
- * const projectsWithMembers = useProjects({ includeMembers: true })
- * const projectsWithoutMembers = useProjects({ includeMembers: false })
+ * const projectsWithMembers = useProjects({includeMembers: true})
+ * const projectsWithoutMembers = useProjects({includeMembers: false})
```
cngonzalez, line 37: maybe worth showing that `useProjects()` can work without any parameters, or both of them

packages/react/src/_exports/sdk-react.ts
```diff
@@ -84,7 +84,7 @@ export {
   type useDocumentProjectionResults,
 } from '../hooks/projection/useDocumentProjection'
 export {useProject} from '../hooks/projects/useProject'
-export {type ProjectWithoutMembers, useProjects} from '../hooks/projects/useProjects'
```
cngonzalez, line 87: Not sure we can remove an export without it being a breaking change I'm afraid :/

cngonzalez (changes requested):
Generally LGTM! mostly stylistic nits, but appreciate the efforts!

packages/react/src/_exports/sdk-react.ts
joshuaellis: Agree – sorry this wasn't meant to be removed!

packages/core/src/projects/projects.test.ts
joshuaellis: can refactor! 👍🏼

packages/core/src/project/project.ts
joshuaellis: ah thats what it does. Great, lets remove, agree we don't need it – should the types disallow it then or is there a reason to be able to provide both at the same time?

cngonzalez: I think historically it's been a bit of a tradeoff between maintaining complex generics / conditional types and letting the occasional unused parameter fall through. There's probably a way to do it elegantly. (personally I think the `scope` parameter is a bit weird now in the era of calling canvas / ML / other resources, and we should just trust `useProjectHostname` but it would be a breaking change now)

cngonzalez (approved):

PR #866, opened 2026-05-06
feat: add organization stores to the sdk

joshuaellis:
### Description

- Adds `organization` and `organizations` stores with `getOrganizationState`, `resolveOrganization`, `getOrganizationsState`, and `resolveOrganizations`
- Introduces conditional `Organization<IncludeMembers>` type so the `members` field is present only when `includeMembers: true`
- Adds `useOrganization` and `useOrganizations` React hooks with suspense support
- Adds type-level tests (`.test-d.ts`) for all new stores and hooks
- Exports new public types: `Organization`, `OrganizationBase`, `OrganizationMember`, `OrganizationsOptions`

### What to review

* Do we have a way to populate other stores from a list store? for example, if someone fetches all the organizations then it would be nice to populate single org stores by their id 🤔 
* Still wondering about branded ID types here, can see how they'd link up more as we grow.

### Fun gif

sorry-ben-stiller-gif-by-apple-tv

### DIFF
--- a/packages/core/src/organization/organization.ts
+++ b/packages/core/src/organization/organization.ts
@@ -1,0 +1,166 @@
+import {switchMap} from 'rxjs'
+
+import {getClientState} from '../client/clientStore'
+import {type SanityInstance} from '../store/createSanityInstance'
+import {type StateSource} from '../store/createStateSourceAction'
+import {createFetcherStore} from '../utils/createFetcherStore'
+
+const API_VERSION = 'v2025-02-19'
+
+/** @public */
+export interface OrganizationMember {
+  sanityUserId: string
+  isCurrentUser: boolean
+  user: {
+    id: string
+    displayName: string
+    familyName: string
+    givenName: string
+    middleName: string | null
+    imageUrl: string | null
+    email: string
+    loginProvider: string
+  }
+  roles: Array<{
+    name: string
+    title: string
+    description?: string
+  }>
+}
+
+/**
+ * The base fields returned from `/organizations/<id>` for every organization.
+ * @public
+ */
+export interface OrganizationBase {
+  id: string
+  name: string
+  slug: string | null
+  createdAt: string
+  createdByUserId: string
+  updatedAt: string
+  deletedAt: string | null
+  dashboardStatus: 'enabled' | 'disabled'
+  aiFeaturesStatus: 'enabled' | 'disabled'
+  mediaLibraryStatus: 'enabled' | 'disabled'
+  requestAccessStatus: 'allowed' | 'disabled'
+  telemetryConsentStatus: 'allowed' | 'msa_denied' | 'customer_denied'
+  oauthAppsStatus: 'allowed' | 'blocked'
+  defaultRoleName: string
+  domains: string[] | null
+}
+
+/** @public */
+export interface OrganizationOptions<
+  IncludeMembers extends boolean = false,
+  IncludeFeatures extends boolean = false,
+> {
+  includeMembers?: IncludeMembers
+  includeFeatures?: IncludeFeatures
+  organizationId: string
+}
+
+/**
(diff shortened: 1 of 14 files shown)


cngonzalez (commented):
To answer this question before I review:

> Do we have a way to populate other stores from a list store? for example, if someone fetches all the organizations then it would be nice to populate single org stores by their id 🤔

I think it's a bit tough because most of the stores are essentially "cache-by-query"; most  of them don't really decompose or sort their results. That's partly by design; it's how we pay attention to subscribers and whether or not we should bother re-fetching that query. That becomes harder if one consumer queries "all organizations" and those become individual organization entries; we'd have to track that that "all organizations" consumer is observing all those entries.

There could be a performance benefit that makes that complexity worth it but I suppose we need to establish a baseline first.

joshuaellis:
> To answer this question before I review:
> 
> > Do we have a way to populate other stores from a list store? for example, if someone fetches all the organizations then it would be nice to populate single org stores by their id 🤔
> 
> I think it's a bit tough because most of the stores are essentially "cache-by-query"; most of them don't really decompose or sort their results. That's partly by design; it's how we pay attention to subscribers and whether or not we should bother re-fetching that query. That becomes harder if one consumer queries "all organizations" and those become individual organization entries; we'd have to track that that "all organizations" consumer is observing all those entries.
> 
> There could be a performance benefit that makes that complexity worth it but I suppose we need to establish a baseline first.

Thanks agree its no easy feat. I was "just wondering" if we had thought about it for the future, maybe not needed now – suppose i was thinking more about seeding the store rather than keeping it observed, but i've not given it enough thought, maybe in reality we should be able to listen to global apis first.

packages/core/src/organization/organization.ts
```diff
@@ -0,0 +1,166 @@
...
+const organization = createFetcherStore({
+  name: 'Organization',
+  getKey: getOrganizationCacheKey,
+  fetcher: (instance) => (options?: OrganizationOptions<boolean, boolean>) => {
+    const organizationId = resolveOrganizationId(options)
+
+    return getClientState(instance, {
+      apiVersion: API_VERSION,
+      scope: 'global',
+      requestTagPrefix: 'sanity.sdk.organization',
```
ryanbonial: [blocking] Please use a tag instead of a prefix. More info and details here: https://github.com/sanity-io/sdk/blob/main/.cursor/rules/request-tag-conventions.mdc

packages/core/src/organizations/organizations.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+const organizations = createFetcherStore({
+  name: 'Organizations',
+  getKey: getOrganizationsCacheKey,
+  fetcher: (instance) => (options?: OrganizationsOptions<boolean, boolean>) => {
+    return getClientState(instance, {
+      apiVersion: API_VERSION,
+      scope: 'global',
+      requestTagPrefix: 'sanity.sdk.organization',
```
ryanbonial: tag fixes here too please. Also, this tag should be `organizations` instead of `organization`

ryanbonial (changes requested):
The logic looks good, but some request tagging changes should be made.

packages/core/src/organization/organization.ts
joshuaellis: Will change, i would have thought we'd use `tag` but in `project` before I changed it it was using `requestTagPrefix` should we do a follow up to tidy these up?

joshuaellis: I actually understand why it was there and why I should have changed prior – will do a follow up!

packages/core/src/organizations/organizations.ts
joshuaellis: changed.

ryanbonial (approved):

PR #868, opened 2026-05-07
fix(core): use tag instead of requestTagPrefix on project calls

joshuaellis:
### Description

In #865 we moved from using the `projects` methods on the client to raw request, I omitted adding a `tag` to the request and removing the `requestTagPrefix` as guided in here – https://github.com/sanity-io/sdk/blob/main/.cursor/rules/request-tag-conventions.mdc so heres that fix 👍🏼 

### Fun gif

sorry-michael-caine-gif


ryanbonial (approved):
Thanks!

PR #869, opened 2026-05-07
refactor(document): separate out processActions

cngonzalez:
### Description
I was about to add some additional actions in service of SDK-913, and couldn't bear the idea of making `processActions.ts` even longer. This separates this out to make actions a bit more maintainable.

### What to review
I asked an LLM to separate out the files but I did a vimdiff of every bit of logic and everything looked the same to me. If you'd like to do the same feel free.

### Testing
Tests not required; functionality not changed.


ryanbonial (approved):
Thanks!

PR #894, opened 2026-05-12
fix(auth): keep subscription alive after /users/me fetch error

joshuaellis:
### Problem

When `/users/me` returns a 401 in dashboard mode, the RxJS subscription in `subscribeToStateAndFetchCurrentUser` terminates permanently. Even after `ComlinkTokenRefresh` obtains a fresh token from the parent frame, the dead subscription never re-fetches `/users/me` — leaving the user stuck at `LOGGED_IN` with `currentUser: null`.

### Solution

Move error handling from the subscriber's `error` callback into a `catchError` operator inside `switchMap`, returning `EMPTY`. This keeps the outer subscription alive so new tokens trigger a retry. Matches the established pattern in `usersStore`, `queryStore`, and `favorites`.

### DIFF
--- a/packages/core/src/auth/subscribeToStateAndFetchCurrentUser.test.ts
+++ b/packages/core/src/auth/subscribeToStateAndFetchCurrentUser.test.ts
@@ -112,4 +112,120 @@ describe('subscribeToStateAndFetchCurrentUser', () => {
 
     subscription.unsubscribe()
   })
+
+  it('recovers from a fetch error when a new token is set', () => {
+    const error = new Error('Unauthorized')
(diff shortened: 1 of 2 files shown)


cngonzalez (approved):
the logic in the actual file makes sense to me! I don't know the structure of the auth store well enough to validate if the tests make sense but I don't see the harm, either

joshuaellis:
"recovers from a fetch error when a new token is set" is the most important imo because thats the issue we're running into, and it failed without the fix so i'm feeling fairly confident in it (famous last words)

PR #898, opened 2026-05-14
feat(core): add release actions to document store

cngonzalez:
### Description
This PR adds the core release API actions to the document store. I chose to put them in the document store because there's so much overlap and logic, and releases are, technically, documents. This PR should hopefully complete our coverage of the Actions API.

React hooks and e2e tests will come in a follow-up PR; I wanted to prevent the size of this PR from growing too much.

### What to review
I made a few choices here:
1. This doesn't update `getActiveReleases` or get the releaseStore to be hooked up via the document's optimistic store. I don't think it's fully required, and would, again, make this PR pretty noisy.
2. The Release type was updated to reflect all the things that a release can actually be. This did mean changing the metadata title field to be possibly undefined, but that's in a type marked `beta` and anyone depending on that field to be defined would have run into issues in the real world; consider it more of a bugfix.
3. Some of the naming -- it's a little funky that we have `releaseName` in the releases store and `releaseId` here. also the files being called `releaseCreate` etc. This is more to match the Actions API, which take a `releaseId` (not a `documentId`) and has actions like `release.create`.
4. Decided to fully error out if an actions API change and mutation (via liveEdit) were in the same batch; if a release document got caught in there, it would result in a no-op and be a bug for the user. Again, this is a pretty rare (possibly impossible) occurrence.

### Testing
Vitest tests were added for the new actions and the new utils.

### DIFF
--- a/packages/core/src/document/actions.ts
+++ b/packages/core/src/document/actions.ts
@@ -3,7 +3,11 @@ import {type PatchMutation as SanityMutatePatchMutation} from '@sanity/mutate/_u
 import {type PatchMutation, type PatchOperations} from '@sanity/types'
 import {type SanityDocument} from 'groq'
 
-import {type DocumentHandle, type DocumentTypeHandle} from '../config/sanityConfig'
+import {
+  type DocumentHandle,
+  type DocumentTypeHandle,
+  type ReleaseHandle,
+} from '../config/sanityConfig'
 import {getEffectiveDocumentId} from './util'
 
 const isSanityMutatePatch = (value: unknown): value is SanityMutatePatchMutation => {
@@ -119,6 +123,17 @@ export type DocumentAction<
   | PublishDocumentAction<TDocumentType, TDataset, TProjectId>
   | UnpublishDocumentAction<TDocumentType, TDataset, TProjectId>
   | DiscardDocumentAction<TDocumentType, TDataset, TProjectId>
+
+/**
+ * Union of every action accepted by `applyDocumentActions` — both document-
+ * level actions and release-lifecycle actions.
+ * @beta
+ */
+export type Action<
+  TDocumentType extends string = string,
+  TDataset extends string = string,
+  TProjectId extends string = string,
+> = DocumentAction<TDocumentType, TDataset, TProjectId> | ReleaseAction<TDataset, TProjectId>
 
 /**
  * Creates a `CreateDocumentAction` object.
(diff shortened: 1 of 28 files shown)


packages/core/src/document/processActions/releaseCreate.ts
```diff
@@ -0,0 +1,59 @@
...
+    name: action.releaseId,
+    state: 'active',
+    metadata: action.metadata,
+  }
+  const mutations: Mutation[] = [{create: releaseDoc}]
+
+  base = processMutations({documents: base, transactionId, mutations, timestamp})
+  working = processMutations({documents: working, transactionId, mutations, timestamp})
+
+  if (!checkGrant(grants.create, working[releaseDocumentId] as SanityDocument)) {
```
ryanbonial, line 43: nitpick: I'm not sure this cast is needed
```suggestion
  if (!checkGrant(grants.create, working[releaseDocumentId])) {
```

packages/core/src/document/actions.ts
```diff
@@ -316,3 +331,194 @@ export function discardDocument<
...
+/**
+ * The release metadata shape accepted by the create/edit release actions.
+ * Mirrors the server-side release document metadata.
+ * @beta
+ */
+export interface ReleaseActionMetadata {
+  title?: string
+  description?: string
+  intendedPublishAt?: string
+  releaseType?: 'asap' | 'scheduled' | 'undecided'
```
ryanbonial: [question] Claude mentioned this (in a way), but why would `releaseType` be optional? If not, then we should provide a default (I assume `undecided`)

ryanbonial (changes requested):
All looks good except for one bug where it is possible for an undefined `metadata` bug

packages/core/src/document/actions.ts
cngonzalez: Went down a rabbit hole on this one! Some background context:
1. The metadata doesn't really affect the release or any behavior, it's mostly for UX
2. It's technically possible to create a release without metadata, it's also part of the client signature https://github.com/sanity-io/client/blob/df7f14ed3857f1f62c27940d6d291842e384dc4f/src/types.ts#L785
3. But the client also has metadata as required for the exported `ReleaseDocument` https://github.com/sanity-io/client/blob/df7f14ed3857f1f62c27940d6d291842e384dc4f/src/types.ts#L1692

I'm not sure if there's some serverside resolution that's missing because in my testing my metadata-less release document is sitting pretty https://ppsg7ml5.api.sanity.io/v2026-05-20/data/query/test?query=*%5B_id+%3D%3D+'_.releases.carolina-test-1'%5D&perspective=drafts. Anyway to protect my sanity I coalesced with the ReleaseDocument but will explore what the deal is later.

packages/core/src/document/processActions/releaseCreate.ts
cngonzalez: Unfortunately `working[releaseId]` could technically be falsey (in this case it never can be since we just set it above) so it's required. Can do a `working[releaseDocumentId]!` if preferred? (this is also copied from the regular `create` document action) 😅

ryanbonial (approved):

PR #905, opened 2026-05-18
feat(core): enable production telemetry on Sanity-controlled domains

ryanbonial:
### Description

Production telemetry is now enabled for SDK apps deployed to Sanity-controlled domains. The old gate only allowed `localhost` / `127.0.0.1`

End users on Sanity-controlled hosts (`*.sanity.studio`, `*.sanity.io`) are authenticated Sanity users with Populus consent records, the same population the Studio's `telemetry-sink` already covers. Apps on customer-controlled domains stay out of the allowlist and skip telemetry entirely.

Changes:

- `getTelemetryEnvironment()` replaces `isDevMode()`. Returns `'development' | 'production' | null` based on hostname (browser) or `NODE_ENV` (node). Customer-controlled hosts return `null` and short-circuit the entire telemetry pipeline.
- Telemetry events renamed from the `SDK Dev *` family to `SDK *` and the resolved environment moved into the event context envelope. Lets dbt key on the new dimension instead of duplicating event families.
- Telemetry stays lazy-loaded. Only the hostname check runs at boot for non-eligible apps; the manager + client modules are still dynamically imported.

While writing tests for the auth-token wait path I also caught a TDZ race that pre-dates this work. The token observable is `shareReplay({bufferSize: 1, refCount: true})`, so a buffered value can be delivered synchronously inside `.subscribe(cb)`, before `sub` has been bound. The original callback did `sub.unsubscribe()` immediately, which threw `ReferenceError`. The Promise rejected and the outer `.catch` silently aborted telemetry init for that session, with no log surface and no retry. The fix gates the callback on a `received` flag and lets a post-subscribe block handle the sync-emission unsubscribe.

Implements SDK-1395.

### What to review

Code:

- `packages/core/src/telemetry/environment.ts` — the allowlist logic. Suffix matching uses leading-dot patterns (`.sanity.studio`, etc.) so apex / lookalike hosts don't match. Test cases for `evilsanity.studio` and apex `sanity.studio` are in the test file.
- `packages/core/src/telemetry/initTelemetry.ts` — the gate change and the token-wait fix described above. The `received`-flag pattern is the smallest delta that handles both sync and async emission cases without changing the dispose-cleanup contract.
- `packages/core/src/telemetry/telemetryManager.ts` — `environment` is now a required option and is recorded in the enriched batch context.
- `packages/core/src/telemetry/events.ts` (renamed from `__telemetry__/sdk.telemetry.ts`) — event name + description changes.

Out-of-band work this depends on:

- Confirmation from the data team that the Populus opt-out rate makes the expanded scope worth shipping. The Tasks section in the Linear ticket has this as a blocker.
- dbt model update for the renamed events (`SDK Session Started`, etc.) and the new `environment` context field.
- Privacy/consent docs update for the wider scope.

Out of scope:

- SDK apps deployed to customer-controlled domains (e.g. `myapp.customer.com`). End users there aren't necessarily authenticated Sanity users with consent records, which is a separate privacy conversation.
- Server-side telemetry. Node runtimes only emit when `NODE_ENV === 'development'`; there's no production-server gate, since the browser-authenticated-user assumption doesn't hold there.

### Testing

- 48 telemetry-package tests pass (`packages/core/src/telemetry/*.test.ts`), 6 of which are new and exercise the auth-token wait path including the sync-emission case that exposed the TDZ bug. The sync-emission test fails against the pre-fix code with a swallowed `ReferenceError`.
- Full core suite: 926/926 pass. Full repo: 1238/1238 pass.
- `pnpm tsc --noEmit` clean for `packages/core`. `pnpm lint` clean (pre-existing TODO and react-refresh warnings only).
- Coverage on the new files is 100% for `environment.ts` and `events.ts`. `initTelemetry.ts` moved from 89.9% → 94.7% statements and 82.5% → 90.9% branches with the new tests.
- The Sanity-domain allowlist is verified with positive cases for each suffix (studio / io / work / dev), a case-insensitivity test, an empty-hostname test, and negative cases for customer domains and lookalike hosts.

### Fun gif
moar


packages/core/src/telemetry/environment.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Apex matches (`sanity.io`) are intentionally excluded; we only emit
+ * from real app surfaces (subdomains).
+ *
+ * @internal
+ */
+const SANITY_CONTROLLED_HOST_SUFFIXES = [
+  '.sanity.studio',
+  '.sanity.io',
+  '.sanity.work',
```
cngonzalez: Any reason for including staging / internal dev hosts? i don't think sanity employees submit telemetry

packages/core/src/telemetry/telemetryManager.ts
```diff
@@ -97,18 +97,24 @@ export function createTelemetryManager(options: TelemetryManagerOptions): Teleme
...
-      context: {
-        version: CORE_SDK_VERSION,
-        environment: 'development' as const,
-        origin: typeof window !== 'undefined' ? window.location.origin : 'node',
-      },
-    }))
+    batch.map((event) => {
+      // Trace events arrive with their own `context`; merge so we don't
+      // drop it. SDK-owned fields below win on conflict.
+      const existing = (event as {context?: Record<string, unknown>}).context
```
cngonzalez: should this be typed as TelemetryEvent?

cngonzalez (dismissed):
tiny nits but everything seems fine

packages/core/src/telemetry/environment.ts
ryanbonial: I thought it would be useful for testing, but you're right, it will likely just be noise.

packages/core/src/telemetry/telemetryManager.ts
ryanbonial: It is, however only the "trace" events have `context` so there is some odd looking narrowing here. I added some comment clarity.

cngonzalez (approved):

PR #926, opened 2026-05-26
ci(e2e): enable turbo remote cache and move e2e-only secrets out of globalEnv

ryanbonial:
### Description
The `End-to-End Tests` workflow was rebuilding everything from scratch for each browser in the matrix because it never set `TURBO_TOKEN` / `TURBO_TEAM`. Turbo logged `Remote caching disabled` and reported `0 cached, 5 total` per job, even though the same package builds had just been produced and uploaded to the remote cache by the `Build` workflow seconds earlier. On a recent run that was roughly 45s of duplicated build work for every browser (chromium / firefox / webkit).
Three small changes:
1. **Enable turbo remote caching in `e2e.yml`.** Add `TURBO_TOKEN` / `TURBO_TEAM` to the env block of both the `setup` job and the `playwright-tests` matrix job, matching what `build.yml`, `test.yml`, `lint.yml`, etc. already do.
2. **Move `SDK_E2E_*` from `globalEnv` onto the one task that actually bakes them in.** The dataset names are only consumed at build time by `apps/kitchensink-react/vite.config.mjs`'s `define` block, so they only need to be part of `@sanity/kitchensink-react#build`'s cache key. Leaving them in `globalEnv` was busting the cache key of every other task whenever a per-browser dataset name changed.
3. **Move `RECAPTCHA_E2E_STAGING_KEY` and `VERCEL_AUTOMATION_BYPASS_SECRET` from `globalEnv` onto the `cleanup` task.** These two are only set in the e2e workflow (never in `Build`), so keeping them in `globalEnv` made every package build hash differ between workflows and prevented cross-workflow cache hits. `test:e2e` already declared them; `cleanup` needed to as well because it transitively imports `getE2EEnv()`, which validates the full env on module load and was crashing in strict env mode without them.
### What to review
- `.github/workflows/e2e.yml`: confirm `TURBO_TOKEN` / `TURBO_TEAM` land on both jobs and don't shadow anything we need from secrets.
- `turbo.json`:
  - New `@sanity/kitchensink-react#build` override carries the right `dependsOn` / `outputs` (turbo overrides replace rather than merge).
  - `cleanup` task now declares all three env groups it needs (`SDK_E2E_*`, `RECAPTCHA_E2E_STAGING_KEY`, `VERCEL_AUTOMATION_BYPASS_SECRET`).
  - Nothing else in the repo was relying on those vars being globally hashed.

### Fun gif

turbo

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -21,6 +21,9 @@ jobs:
   # shared setup: install deps and warm Playwright browser cache for downstream jobs
   setup:
     runs-on: ubuntu-latest
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+      TURBO_TEAM: ${{ vars.TURBO_TEAM }}
     steps:
       - name: 📥 Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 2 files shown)


cngonzalez (approved):
dope, thank you!

gu-stav (approved):

PR #931, opened 2026-06-04
feat(ailf): add AILF

cngonzalez:
### Description
This PR scaffolds AILF runs in CI and adds an example task to build off of. The goal here is not to create a full suite, but rather to make it possible for folks to easily add their own tasks and start a discussion of the kinds of prompts we want to test and what kinds of docs we want to run tests against.

### What to review
I made some choices here:
1. The initial test doesn't use the reference documentation. I need to verify if the ability to read them was added.
2. AILF won't automatically run on every PR; just those that touch AILF tasks. Devs can add the appropriate trigger if they want to include a run.
3. The AILF job is scheduled to run every Thursday via Github actions.

### Testing
Many previous runs seem to work fine. To see an example, follow the new Github action to the AILF report dashboard: https://github.com/sanity-io/sdk/actions/runs/27144381418?pr=931

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,24 +15,23 @@ RECAPTCHA_E2E_STAGING_KEY=
 # This can be found in the "Deployment Protection" settings section of the sanity-io Vercel project.
 VERCEL_AUTOMATION_BYPASS_SECRET=
 
-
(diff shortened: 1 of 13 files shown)


ryanbonial (approved):
Looks good 👍

PR #932, opened 2026-06-08
fix(document): pass identity to local acl evaluations

cngonzalez:
### Description
Canvas and some user-defined roles use the `$identity` or `identity()` parameter in their grant definitions, where `$identity` refers to the current user. There previously no concept of this in the document store. This PR adds a subscription to the current user and injects it in those places in the grants evaluation so that we can accurately pass those filters.

(Note that this only affected the optimistic store calculations, but those are required to work correctly without breaking the whole document store).

### What to review
Any cases I'm missing? Is there a better way?

### Testing
Tests were added.


binoy14 (approved):
Looks good to me as far as I can tell

PR #934, opened 2026-06-09
fix(core): ignore unknown grants in createGrantsLookup

binoy14:
## Problem

`createGrantsLookup` crashes when the dataset ACL contains a permission that isn't modeled by the `Grant` type. The real ACL API can return values like `"manage"`, e.g. `["history", "read", "update", "create", "manage"]`.

The function indexes into a fixed record keyed by the four known grants:

```ts
const set = filtersByGrant[grant] // undefined when grant === "manage"
set.add(entry.filter)             // 💥 TypeError: Cannot read properties of undefined (reading 'add')
```

## Fix

Skip grants we don't model (`if (!set) continue`). Known grants still resolve correctly.

## Test

Added `should ignore unknown permissions like "manage"` to `permissions.test.ts`, which reproduced the crash before the fix and passes after.

> Note: `DatasetAcl.permissions` is typed as `Grant[]`, so the test casts the array. The real ACL API returning `"manage"` is the type-level mismatch that made this reachable in practice.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/document/permissions.test.ts
+++ b/packages/core/src/document/permissions.test.ts
@@ -52,6 +52,18 @@ describe('createGrantsLookup', () => {
       expect(evaluateSync(grants[key], {params: {document: dummyDoc}}).data).toBe(true)
     })
   })
+
+  it('should ignore unknown permissions like "manage"', () => {
+    const datasetAcl = [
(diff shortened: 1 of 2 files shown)


cngonzalez (approved):
LGTM!

PR #936, opened 2026-06-09
fix(deps): unpin tsx via catalog now that Node 24 regression is resolved

binoy14:
Reverts the `tsx <4.22.0` pin added in #922 to dodge a Node 24
regression in tsx's `?tsx-namespace=` query propagation, which broke
`@sanity/pkg-utils`' `pkg build` (it loads `package.config.ts` via
`tsImport` from `tsx/esm/api`).

tsx 4.22.4 (privatenumber/tsx#803, "resolve CommonJS directory requires
inside dependencies") fixes the surface the SDK hits. Verified locally
on Node 24.13.0: clean `core`+`react` `pkg build` runs pass on 4.22.4.

Moves the version into a pnpm `catalog` so the root and `@repo/e2e`
share one source of truth instead of two literal specifiers, and drops
the now-unnecessary `pnpm.overrides` ceiling.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "prettier": "^3.8.3",
     "read-package-up": "^12.0.0",
     "rimraf": "^6.0.1",
-    "tsx": "<4.22.0",
+    "tsx": "catalog:",
     "turbo": "^2.9.14",
     "typedoc": "^0.28.19",
     "typescript": "^5.9.3",
@@ -84,8 +84,7 @@
     "overrides": {
       "@babel/types": "7.29.7",
       "basic-ftp@<5.3.1": "^5.3.1",
-      "form-data@<2.5.4": "^2.5.4",
-      "tsx": "<4.22.0"
+      "form-data@<2.5.4": "^2.5.4"
     }
   }
 }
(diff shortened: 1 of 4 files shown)


binoy14:

- #936 ◀

binoy14:
Closing — the premise ("Node 24 regression is resolved") is incorrect. tsx 4.22.4 still breaks `@sanity/pkg-utils`' `pkg build` on Node 24.

The `build (24)` CI job failed with:
```
Error: ENOENT: no such file or directory, open 'node:os?tsx-namespace=...'
  at node:internal/modules/cjs/loader ...
  at tsx/dist/register-...mjs (the .js transformer hook)
```

Root cause is a Node *patch*-specific regression: tsx 4.22.4 passes on Node 24.13.0 (what local was running) but fails on Node 24.16.0 (what `actions/setup-node@node-version:24` resolves to, and what `.nvmrc` pins). tsx's CJS loader hook appends `?tsx-namespace=` to a `node:` built-in specifier, which the Node 24.16 CJS loader then tries to `readFileSync`.

This is tracked upstream and still **open**: privatenumber/tsx#801 (repro is literally `@sanity/pkg-utils` on Node 24). The `<4.22.0` pin should stay until #801 is closed and released.

PR #950, opened 2026-06-10
chore(deps): update @sanity/pkg-utils to v10

binoy14:
### Description

Updates `@sanity/pkg-utils` from `^9.2.3` to `^10.5.3` (latest major).

- **tsconfig presets moved out of pkg-utils.** v10 relocated its tsconfig presets to the new `@sanity/tsconfig` package, so `@repo/tsconfig` now depends on `@sanity/tsconfig` (catalog `^2.1.0`) and `base.json` extends `@sanity/tsconfig/strictest` instead of `@sanity/pkg-utils/tsconfig/strictest.json`.
- **Removed the `tsx <4.22.0` pin.** The pin (and the `pnpm.overrides` entry) existed to dodge a Node 24.16 regression in pkg-utils **9.x**'s `pkg build`. pkg-utils **10** builds cleanly on Node 24.16 + tsx 4.22.4, so the pin is obsolete. The catalog `tsx` is now `^4.22.4`.
- **Unified tsx to a single version.** Removing only the override (while the catalog stayed pinned `<4.22.0`) left two tsx versions in the tree. Because `vite` declares `tsx` as a peer, pnpm built two type-incompatible copies of `vite`, which broke `apps/kitchensink-react` typecheck (`TS2322` on vite's private `_pluginContextMap`). Unifying tsx collapses vite back to one instance.

The other two v10 breaking changes (removed internal API exports; removed `noPackageJsonMain`/`noPackageJsonModule` lint rules) don't apply to this repo.

### What to review

- `pnpm-workspace.yaml` — `@sanity/pkg-utils` bump, new `@sanity/tsconfig` catalog entry, `tsx` unpinned to `^4.22.4`.
- `packages/@repo/tsconfig/{package.json,base.json}` — new `@sanity/tsconfig` dep and updated `extends`.
- `package.json` — removed the `pnpm.overrides` `tsx` entry.
- `pnpm-lock.yaml` — regenerated.

### Testing

- `pnpm build` — **5/5 tasks green** (core, react, kitchensink-react, etc.) on Node **24.16.0** (matches `.nvmrc`, the version that previously surfaced the tsx regression — so this is a true-green, not a false one from an older Node patch).
- `packages/core` and `packages/react` `pkg build --strict --clean --check` both pass.
- `apps/kitchensink-react` `tsc && vite build` passes (the previously failing typecheck).

> [!NOTE]
> Watch the CI `build (24)` job — a prior unpin attempt (PR #936) failed there on pkg-utils 9.x. This PR is on v10, and local Node here equals CI's (24.16.0), so it should match.

### Fun gif

ship it


binoy14:

- #950 ◀

cngonzalez (approved):
LGTM!

PR #952, opened 2026-06-10
chore(kitchensink): run typegen via @sanity/codegen, drop @sanity/cli-v3 bin

binoy14:
### Description

The kitchensink's `schema:extract` and `types:generate` scripts ran through the experimental `@sanity/cli-v3` bin (`npm:@sanity/cli@3.88.1-typegen-experimental.0`). That bin can no longer be invoked in this repo: booting a v3 `@sanity/cli` eagerly loads the v5 `sanity` package's command bridge (`sanity/_internal`), whose `cliProjectCommands.commands` is a `Proxy` that deliberately throws an "installed version of `@sanity/cli` is not compatible with the installed version of `sanity`" error. So any `@sanity/cli-v3/bin/sanity …` call crashes before the command even runs.

The experimental multi-schema/projections typegen feature lives entirely in `@sanity/codegen` (+ its bundled `groq`), not in the CLI framework — `sanity typegen generate` is just a thin wrapper that reads `sanity-typegen.json` and drives a worker built on `@sanity/codegen`. This PR calls that library directly and drops the unusable bin:

- **`scripts/typegen.mjs`** (new) — a faithful port of `@sanity/cli`'s `workers/typegenGenerate.ts` + `actions/typegen/generateAction.ts`. Reads `sanity-typegen.json` and runs `@sanity/codegen`'s `TypeGenerator`. Never imports `sanity/_internal`, so the version trap never fires.
- **`types:generate`** → `node scripts/typegen.mjs` (was the `@sanity/cli-v3` bin).
- **`schema:extract`** → the studio's matching `sanity` CLI (was the `@sanity/cli-v3` bin). Extract must load `sanity.config.ts`, so it has to use the CLI that matches the installed `sanity`.
- **deps**: replace the `@sanity/cli-v3` alias with a direct `@sanity/codegen: 3.88.1-typegen-experimental.0` dependency — the real package that does the work. This also drops cli-v3's large transitive tree from the lockfile.

This is prep for Studio v6 (ships with vite 8). It is intended as the base of a stack, with the vite 8 upgrade to follow on top.

### What to review

- `scripts/typegen.mjs` — does it faithfully mirror the upstream typegen worker/action (schema loading, query declaration streaming, prettier formatting)?
- The `require('@sanity/codegen')` (vs a top-level `import`): the experimental build's ESM entry (`lib/index.mjs`) is mis-built — it calls `createRequire(__filename)`, and `__filename` is undefined in ESM scope, so a static import throws on load. `require` selects the working CJS entry. There's an inline comment documenting this.
- `package.json` script + dependency changes.

### Testing

- `pnpm --filter @sanity/kitchensink-react types:generate` produces a `sanity.types.ts` that is **byte-for-byte identical** to the output of the old `@sanity/cli-v3` bin (verified via `git diff` on the generated file — no drift).
- The multi-schema/projections feature is preserved: generated types are tagged per schema (`SchemaOrigin<…, 'ppsg7ml5.test'>` / `'vo1ysemo.production'`) and import `ProjectionBase`/`PickSchema`/`SchemaOrigin` from the experimental `groq`.
- `sanity schema extract --workspace …` was verified to extract both workspace schemas (validated against `sanity@5.31.1`); the extracted JSON re-feeds the typegen script with no output drift.

> Note: the generated `schema.*.json` / `sanity.types.ts` are gitignored and produced by the `install` hook, so they are not part of this diff.

### Fun gif

Cute kittens


binoy14:

- #952 ◀
    - #954

cngonzalez (dismissed):
dope, good stuff

cngonzalez (approved):

PR #954, opened 2026-06-10
chore(kitchensink): build via App SDK CLI and auto-expose e2e env vars

binoy14:
### Description

Runs the kitchensink through the **Sanity App SDK CLI** instead of a hand-maintained raw-Vite setup.

- **Build/dev via the App SDK CLI**: `build` → `sanity build`, `preview` → `sanity preview` (`dev`/`deploy` already used `sanity`). The old `vite.config.mjs` is inlined into `sanity.cli.ts`'s `vite` hook, using the CLI's native `reactCompiler: {target: '19'}`. Dropped `@vitejs/plugin-react`, `vite.config.mjs`, `index.html`, and `src/main.tsx`.
- **e2e env vars auto-exposed**: non-secret e2e vars renamed `SDK_E2E_*` → `SANITY_APP_E2E_*` so the App SDK auto-exposes them on `import.meta.env` (no manual `define`/`loadEnv`). Secrets stay `SDK_E2E_*` and server-only. A `SANITY_APP_E2E_MODE` flag now drives `isE2E`.
- **All e2e (incl. webkit) runs through `sanity dev`**: dropped the standalone raw-vite webServer path, so the extra vite `server` config and the root-level `vite` devDependency are no longer needed (vitest resolves vite transitively).

> **Note:** this branch no longer contains the Vite 8 upgrade — that was split out (catalog stays on Vite 7). The branch name is now a misnomer but kept for continuity.

### What to review

- `sanity.cli.ts` — inlined Vite config + native `reactCompiler`.
- The `SDK_E2E_*` → `SANITY_APP_E2E_*` rename across `@repo/e2e`, `e2e.yml`, `turbo.json`, and the app's `import.meta.env` reads — and that secrets were **not** renamed (must not reach the client bundle).
- `playwright.config.ts` — webkit now uses `sanity dev` like every other project.

### Testing

- `pnpm --filter @sanity/kitchensink-react lint` ✅, `tsc --noEmit` ✅, `sanity build` ✅, `pnpm knip` ✅.
- vitest runs without root `vite` declared (resolves it transitively).
- Verified `SANITY_APP_E2E_*` auto-exposes in the bundle and `envDir` reads the root `.env`.

### Fun gif

Cute kittens


binoy14:

- #952
    - #954 ◀

apps/kitchensink-react/index.html
```diff
@@ -1,14 +0,0 @@
-<!doctype html>
-<html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
-    <link rel="stylesheet" href="./src/global.css" />
```
ryanbonial, line 6: I think we may need to keep this global.css (import it elsewhere) and maybe delete the favicon

ryanbonial (changes requested):
I think that the global.css got lost a while ago, but it would be nice to add it back in as an import in App.tsx 🙏

apps/kitchensink-react/index.html
binoy14: good catch, fixed

ryanbonial (approved):

PR #967, opened 2026-06-12
feat(ailf): add docs improvement agent job

cngonzalez:
### Description
This PR adds a script that uses Claude and the Sanity MCP to respond to AILF feedback. This is mostly intended to be used in CI; you'll see the AILF Github workflow updated so that we can get feedback during AILF runs. This is only the first part of the full workflow -- we can follow this up with using an agent to write a content release that actually reflects this feedback.

While I was here I also removed some things that led to constant AILF report noise. We (purposely) don't encourage `useDocumentPreview` and adding the `best-practices` document also meant that the report focused on things that aren't actually in our docs (we kept getting false dings on not documenting `Suspense` for example).

I also removed llm-ese from places, just to keep things a bit terser.

### What to review
The script itself is hopefully rather defensive; we've excluded most paths to writing anything. Let me know if there are places I can harden up.

### Testing
You can see the results of the output in the AILF CI job on this PR!

### DIFF
--- a/packages/@repo/ailf-eval/README.md
+++ b/packages/@repo/ailf-eval/README.md
@@ -6,10 +6,10 @@ AILF evaluates how well the SDK docs enable AI coding tools to implement feature
 
 ## Layout
 
-- `.ailf/ailf.config.ts` — project configuration (authored with `defineRepoConfig` from `@sanity/ailf`): the docs source (Sanity project/dataset/site URL) and the `taskSource` (set to `repo` so AILF reads the tasks in this package). This is the only config file `ailf run` reads.
-- `.ailf/tasks/*.task.ts` — task definitions, authored with `defineTask` from [`@sanity/ailf`](https://www.npmjs.com/package/@sanity/ailf). Each task evaluates whether an AI coding agent can implement a feature using the docs as context. `sdk-list-documents.task.ts` is the starter task: it checks that an agent can build a paginated document list with the App SDK's `useDocuments` hook.
+- `.ailf/ailf.config.ts` — project configuration (authored with `defineRepoConfig` from `@sanity/ailf`): the docs source (Sanity project/dataset/site URL) and the `taskSource` (set to `repo` so AILF reads the tasks in this package).
+- `.ailf/tasks/*.task.ts` — task definitions, authored with `defineTask` from [`@sanity/ailf`](https://www.npmjs.com/package/@sanity/ailf), with a `defineSdkTask` which also injects the API reference docs. Each task evaluates whether an AI coding agent can implement a feature using the docs as context. `sdk-list-documents.task.ts` is the starter task: it checks that an agent can build a paginated document list with the App SDK's `useDocuments` hook.
 
-The workflow that runs evaluations lives at [`.github/workflows/ailf-eval.yml`](../../../.github/workflows/ailf-eval.yml) in the repo root.
+The workflow that runs evaluations lives at [`.github/workflows/ailf-eval.yml`](../../../.github/workflows/ailf-eval.yml) in the repo root. After each eval it also runs the docs-improvement job and appends the proposal to the run summary, below the dashboard report link (needs the `SANITY_DOCS_READ_TOKEN` and `ANTHROPIC_API_KEY` repo secrets).
 
 ## Authoring a task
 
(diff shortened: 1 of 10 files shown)


packages/@repo/ailf-eval/README.md
```diff
@@ -19,6 +19,20 @@ Full field reference: https://github.com/sanity-labs/ai-literacy-framework/blob/
 
 ## Scripts
 
-- `pnpm eval` — submit the tasks in this package to the AILF API and write a score report (needs `AILF_API_KEY`). Run from the repo root as `pnpm ailf`.
+- `pnpm eval` — submit the tasks in this package to the AILF API and write a score report (needs `AILF_API_KEY`). Run from the repo root as `pnpm ailf:eval`.
+- `pnpm improve-docs` — run from the repo root as `pnpm ailf:improve-docs`. This will run headless Claude over the latest score report with read-only access to the docs dataset, and print proposed docs fixes as markdown to stdout (also saved next to the report as `docs-improvement.md`).
+  Needs `SANITY_DOCS_READ_TOKEN` (please use a token that doesn't have write access) and `ANTHROPIC_API_KEY` (the agent runs against the Anthropic API). Applying proposals will be the next step, and will likely have a separate read token. See `scripts/improve-docs.ts` for the trust model.
```
ryanbonial, line 24:
```
Applying proposals will be the next step, and will likely have a separate read token.
```
I'm a bit confused here, applying suggestions seems like this should be a "write" token

packages/@repo/ailf-eval/scripts/improve-docs.ts
```diff
@@ -0,0 +1,275 @@
...
+try {
+  const mod = await import(pathToFileURL(CONFIG).href)
+  const src = mod.default?.source ?? {}
+  cfgProject = src.projectId ?? ''
+  cfgDataset = src.dataset ?? ''
+} catch (err) {
+  fail(`could not evaluate ${CONFIG}: ${err}`)
+}
+const PROJECT_ID = cfgProject || fail(`set source.projectId in ${CONFIG}`)
+const DATASET = cfgDataset || fail(`set source.dataset in ${CONFIG}`)
```
ryanbonial, line 82: Re "let me know if there are places I can harden up": you anticipated this in the header comment ("We could add another step to check the token's permissions if deemed necessary"). I think it's worth deeming necessary, since it's one HTTP request and it turns the README's "please use a read-only token" into an enforced invariant. The `/acl` endpoint returns the grants the requesting identity has on the dataset (it's the same endpoint Studio's grants store uses), so we can fail the run before the agent gets a single turn:

```suggestion
const DATASET = cfgDataset || fail(`set source.dataset in ${CONFIG}`)

// Preflight: refuse to hand the agent a token that can write. The /acl
// endpoint returns the grants this identity has on the dataset (the same
// endpoint Studio uses to compute permissions), so a mis-provisioned token
// fails the run before the agent gets a single turn.
const aclRes = await fetch(
  `https://api.sanity.io/v2021-06-07/projects/${PROJECT_ID}/datasets/${DATASET}/acl`,
  {headers: {Authorization: `Bearer ${SANITY_DOCS_READ_TOKEN}`}},
)
if (!aclRes.ok) fail(`could not verify token grants (HTTP ${aclRes.status}); refusing to run`)
const grants: {permissions?: string[]}[] = await aclRes.json()
const WRITE_PERMISSIONS = new Set(['create', 'update', 'editHistory'])
if (grants.some((grant) => grant.permissions?.some((p) => WRITE_PERMISSIONS.has(p)))) {
  fail('SANITY_DOCS_READ_TOKEN has write grants on the docs dataset; use a read-only token')
}
```

Two notes:

- This shouldn't conflict with the planned follow-up where an agent applies the proposals: that job will have its own write-capable token in a separate env var, so it never passes through this check. The guard actually protects that future, because wiring the write token into *this* job by mistake fails loudly instead of silently widening the propose agent's access.
- Worth a quick live check that the endpoint returns grants for robot tokens the same way it does for user sessions before merging (Studio exercises it with user credentials). If it does, this also fails closed: an API error refuses to run rather than proceeding unchecked.

cngonzalez: good feedback -- I've removed the token from our secrets for now. 

Especially with Claude's comment (the Github action comment I mean) I think we actually better limit the token to just published documents (it would be great if MCP didn't require a token at all!), which would be a different grants check. On Monday I'll have more of a think of what that check looks like. I actually think we could use the SDK for this! 

Thanks again!

packages/@repo/ailf-eval/README.md
cngonzalez: Miswrote 😓 Will fix on Monday!

PR #998, opened 2026-06-22
fix(releases): handle missing metadata

cngonzalez:
### Description
The `sortReleases` logic was a bit more fragile than I realized; although we have type guards across the client et al that prevent releases being made without metadata, it's [possible to create a release document without metadata](https://ppsg7ml5.api.sanity.io/v2026-07-08/data/query/test?query=*%5B_id+%3D%3D+%27_.releases.carolina-test-1%27%5D&perspective=drafts). This ended up bricking the whole store and resulted in releases hooks silently returning nothing.

This PR accounts for the possibility of missing metadata and improves error handling.

### What to review
Are the guards still fragile in any way?

### Testing
Tests added and updated. I don't know why Fallow thinks that file is not being used. It is being used.

### DIFF
--- a/apps/kitchensink-react/src/routes/releases/ReleasesAutocomplete.tsx
+++ b/apps/kitchensink-react/src/routes/releases/ReleasesAutocomplete.tsx
@@ -22,7 +22,9 @@ export function ReleasesAutocomplete({
       <Autocomplete
         id="release-autocomplete"
         filterOption={(query, option) =>
-          (option.payload.metadata.title ?? '').toLowerCase().indexOf(query.toLowerCase()) > -1
+          (option.payload.metadata?.title ?? option.payload.name ?? '')
+            .toLowerCase()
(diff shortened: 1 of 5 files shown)


gu-stav:
@cngonzalez is that related to https://github.com/sanity-io/ada/pull/3462? I was wondering at the time whether that should be possible at all, because the types said no, which is why I never merged it, even though its an error in Sentry.

cngonzalez:
> @cngonzalez is that related to [sanity-io/ada#3462](https://github.com/sanity-io/ada/pull/3462)? I was wondering at the time whether that should be possible at all, because the types said no, which is why I never merged it, even though its an error in Sentry.

Hahah my comment there doesn't help either 😛 And yes I think so! The release linked in this PR's description was made with a cURL request I think so it's not super likely, but since there's no backend validation on it (since releases are just Content Lake documents) I think it definitely is possible.

binoy14 (approved):

PR #1015, opened 2026-07-07
fix(core): recover queries after fetch, live-connection, and suspense errors

binoy14:
### Description

A query that errored once could get permanently stuck until page reload. Three distinct defects, each fixed with a regression test:

1. **Errored query keys never refetched.** `catchError(() => EMPTY)` on the per-key `groupBy` group pipe completed the group's subscription; re-added keys emitted into a dead Subject. Moved `tap`/`catchError` inside the per-event fetch stream so an error only ends that fetch.
2. **One live-events error bricked the whole store.** Non-CORS errors from `client.live.events()` set the store-global `state.error`, which every selector throws and nothing clears. Now retried with a delay; `DisconnectError` ends live updates quietly instead of reconnecting against the server's instruction.
3. **A query that errored while suspended left an immortal key.** The component never commits, so no subscriber exists and nothing could remove the errored key — every retry rethrew the stored error. `resolveQuery` now holds a temporary subscriber, released after the clear delay when its promise settles (immediately on abort).

Fixes 1 and 3 depend on each other; 2 is independent. Dead `initializeQuery`/`cancelQuery` reducers removed.

FIXES SDK-1872

### What to review

`packages/core/src/query/queryStore.ts` — the three changes above. Behavior notes: abort removes the key immediately via `removeSubscriber` (was `cancelQuery`); resolved-but-never-subscribed query state now expires after the clear delay instead of persisting forever.

### Testing

Five regression tests in `queryStore.test.ts`, one per behavior (each fails without its fix). Full core suite passes (83 files / 1000 tests). Verified manually: force one fetch failure via DevTools offline, go back online, retry — previously no request ever fired again; now it refetches.

### Fun gif

Try again

### DIFF
--- a/packages/core/src/query/queryStore.test.ts
+++ b/packages/core/src/query/queryStore.test.ts
@@ -7,12 +7,7 @@ import {isCanvasResource} from '../config/sanityConfig'
 import {createSanityInstance, type SanityInstance} from '../store/createSanityInstance'
 import {type StateSource} from '../store/createStateSourceAction'
 import {getQueryState, resolveQuery} from './queryStore'
-
-vi.mock('./queryStoreConstants', async (importOriginal) => ({
-  ...(await importOriginal<typeof import('./queryStoreConstants')>()),
-  QUERY_STATE_CLEAR_DELAY: 10,
-  LIVE_EVENTS_RETRY_DELAY: 10,
-}))
+import {LIVE_EVENTS_RETRY_DELAY, QUERY_STATE_CLEAR_DELAY} from './queryStoreConstants'
 
 vi.mock('../client/clientStore', () => ({
   getClientState: vi.fn(),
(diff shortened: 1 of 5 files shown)


binoy14:

- #1015 ◀

packages/core/src/query/queryStore.test.ts
```diff
@@ -317,6 +325,112 @@ describe('queryStore', () => {
...
+    )
+
+    const query = '*[_type == "movie"]'
+    const state1 = getQueryState(instance, {query})
+    const unsub1 = state1.subscribe()
+    expect(() => state1.getCurrent()).toThrow('transient network failure')
+    unsub1()
+
+    // Wait for the clear delay so the key is fully removed from state
+    await new Promise((resolve) => setTimeout(resolve, 20))
```
cngonzalez: should we use fake timers instead of real timeouts? we don't really have a problem with flaky tests right now so it's not really a big deal (and we probably do this in other test files) just making a note of it

cngonzalez (dismissed):
fix and tests make sense to me, one nit about setTimeout

cngonzalez (approved):

PR #1021, opened 2026-07-09
chore: migrate from eslint-plugin-import to eslint-plugin-import-x

binoy14:
### Description

Migrates from the unmaintained-for-flat-config `eslint-plugin-import` to `eslint-plugin-import-x`.

- Swap `eslint-plugin-import` → `eslint-plugin-import-x@^4.17.1`; rename all `import/*` rules and settings to `import-x/*`.
- Simplify the TS resolver to `createTypeScriptImportResolver({alwaysTryTypes: true})` — the previous `extensions` override broke resolution of extensionless `.tsx` imports (silently disabling `no-cycle` for React components), and the `project` globs were dead config in per-package turbo runs.
- Fix violations surfaced because `no-extraneous-dependencies` was silently inert under the old plugin (it never reported anything): allow `vite.config.mjs` in the shared devDependencies globs and allow root `scripts/**` to import devDependencies.
- Dedupe configs: remove local `ignores` blocks that duplicated the base config, merge kitchensink's two identical `no-extraneous-dependencies` overrides, drop e2e's empty rules block and a `rules-of-hooks` line already set by the preset.

Lint is also ~2.4× faster: `eslint .` on `packages/core` went from ~8.3s to ~3.5s.

### What to review

- `packages/@repo/config-eslint/index.mjs` — the rule/settings rename and resolver change.
- The per-package `eslint.config.mjs` cleanups are behavior-preserving deletions.

### Testing

`pnpm lint` passes across the workspace. Resolver behavior verified by invoking `createTypeScriptImportResolver` directly (`.tsx` now resolves; previously it didn't), and `react-hooks` rules confirmed still at error level via `--print-config`.

### Fun gif

Sweeping up

### DIFF
--- a/apps/kitchensink-react/eslint.config.mjs
+++ b/apps/kitchensink-react/eslint.config.mjs
@@ -5,19 +5,6 @@ import reactConfig from '@repo/config-eslint/react'
 export default [
   {
     ignores: [
-      '.DS_Store',
-      '**/node_modules',
-      '**/build',
-      '**/dist',
-      '.env',
-      '.env.*',
-      '!.env.example',
-
-      // Ignore files for PNPM, NPM and YARN
-      'pnpm-lock.yaml',
-      'package-lock.json',
-      'yarn.lock',
-
       // Ignore files for Sanity TypeGen
       'sanity.types.ts',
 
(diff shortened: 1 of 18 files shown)


binoy14:

- #1021 ◀

cngonzalez (approved):

PR #1023, opened 2026-07-09
refactor(core): break import cycle between query and releases stores

binoy14:
### Description

Removes the import cycle `queryStore → getPerspectiveState → releasesStore → queryStore` that was suppressed with three `import-x/no-cycle` disables (added in #1021).

- New `client/liveEvents.ts`: shared `observeLiveEvents()` extracted from queryStore (CORS/disconnect/retry handling). Both consumers share one EventSource — `@sanity/client` caches the live stream per URL + connection options, so the request tag is now `live-events` (was `query-store`) for both.
- New `releases/observeReleases.ts`: releasesStore fetches `releases::all()` directly via clientStore instead of through queryStore, breaking the cycle. Live invalidation semantics (mid-fetch messages, sync-tag matching, initial synchronous emission) mirror the old queryStore path.
- `getPerspectiveState` now binds at module scope; the lazy-binding workaround for circular init is no longer needed.
- All three `eslint-disable-next-line import-x/no-cycle` comments removed; the rule now passes at error level.

### What to review

- The RxJS in `observeReleases.ts`: mid-flight live messages are re-evaluated once sync tags land (combineLatest + BehaviorSubject), with `distinctUntilChanged` on the event id preventing refetch loops.
- `startWith(undefined)` in `observeReleases` preserves the synchronous initial emission the store previously got from `StateSource`, so the releases hooks don't newly suspend.
- Observability note: dashboards filtering on the `sanity.sdk.query-store` live-connection tag should switch to `sanity.sdk.live-events`.

### Testing

New unit tests for `observeReleases` (initial emission, sync-tag refetch with `lastLiveEventId`, non-matching tags, mid-flight message re-evaluation) and `liveEvents` (message filtering, CORS callback, DisconnectError, retry), plus CORS-path tests for both stores. Existing queryStore live tests still exercise the real extracted code (only clientStore is mocked). 1007 core + 322 react tests pass.

### Fun gif

Cutting the knot

### DIFF
--- a/packages/core/src/client/liveEvents.test.ts
+++ b/packages/core/src/client/liveEvents.test.ts
@@ -1,0 +1,151 @@
+import {
+  ConnectionFailedError,
+  CorsOriginError,
+  DisconnectError,
+  type LiveEvent,
+  type LiveEventMessage,
+  type SanityClient,
+} from '@sanity/client'
+import {of, Subject} from 'rxjs'
+import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
+
+import {createSanityInstance, type SanityInstance} from '../store/createSanityInstance'
+import {type StateSource} from '../store/createStateSourceAction'
+import {getClientState} from './clientStore'
+import {LIVE_EVENTS_RETRY_DELAY, observeLiveEvents} from './liveEvents'
+
+vi.mock('./clientStore', () => ({getClientState: vi.fn()}))
+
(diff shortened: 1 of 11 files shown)


binoy14:

- #1021
    - #1023 ◀

cngonzalez (dismissed):
nice, thanks!

cngonzalez (approved):

PR #1029, opened 2026-07-10
fix(react): ensure explicit "default" resource is used

cngonzalez:
### Description
After all the switching between making the resource changes a breaking v3 change and deciding to incorporate it in v2, there was a bit of confusion of logic. The sole `ResourceProvider` invoked in `SDKProvider` always used the configs passed in `config` and made the first one the `default` to match legacy behavior, but didn't access the new `resources` parameter.

This PR resolves that by making sure both are being read. This change also revealed a regression where resources weren't being passed to the perspective store if you changed perspectives; that's now fixed.

### What to review
The simpler logic chain in `SDKProvider`; any catches or gotchas?

### Testing
Kitchensink was switched to the future format to make sure we're e2e testing it. Some vitest tests were added for the SDKProvider changes; the perspective / query store regression was harder to address.

### DIFF
--- a/packages/react/src/components/SDKProvider.tsx
+++ b/packages/react/src/components/SDKProvider.tsx
@@ -1,4 +1,9 @@
-import {type DocumentResource, isImportError, type SanityConfig} from '@sanity/sdk'
+import {
+  type DocumentResource,
+  isDatasetResource,
+  isImportError,
+  type SanityConfig,
+} from '@sanity/sdk'
 import {type ReactElement, type ReactNode, useEffect, useMemo} from 'react'
 import {ErrorBoundary, type FallbackProps} from 'react-error-boundary'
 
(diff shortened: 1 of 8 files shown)


packages/react/src/components/SDKProvider.tsx
```diff
@@ -72,10 +71,34 @@ export function SDKProvider({
...
   return (
     <ErrorBoundary FallbackComponent={ChunkAwareFallback}>
       <ResetChunkReloadFlagOnMount />
-      <ResourceProvider {...resolvedConfig} fallback={fallback}>
+      {/* Spread the first config so SanityInstance can use the auth,
+        perspective, and other config fields; `resource` lets an explicitly
+        provided `default` resource override the config-derived resource. */}
+      <ResourceProvider
+        {...defaultConfig}
+        resource={resourcesValue[DEFAULT_RESOURCE_NAME]}
```
ryanbonial, line 99: This overrides any resource from the spread defaultConfig. Which I think will cause 2 problems that the Claude code review sort of flagged:

1. Resources-only apps (no `projectId`/`dataset` on config) break `useProject()` and `useDatasets()`. Both read `instance.config.projectId`, which is never set from `resources.default`.

2. When config and resource point at different projects, hooks split. `useProject()` hits config's project (via the instance) while `useQuery()` hits the resource context's project (via `useNormalizedResourceOptions`)

ryanbonial (changes requested):

packages/react/src/components/SDKProvider.tsx
cngonzalez: Ooh, good call. I suppose these are existing bugs now; if someone manually declares a `ResourceProvider` they'll also be surprised since I believe it passes down the existing `SanityInstance` if it exists in the tree.

Since this is a bit outside the scope of this PR, let me make another PR with some e2e / regression testing and address it.

ryanbonial: Agreed the nested `ResourceProvider` case is pre-existing: it reuses the parent instance and only swaps the resource context. The top-level case is new in this PR though. Before, `resources` was ignored, so config and context couldn't diverge at the root.

Deferring the deeper fix is fine. Could we add one small guard here first? When config has no `projectId`/`dataset` and `resources.default` is a dataset resource, copy its `projectId`/`dataset` into the config passed to `ResourceProvider`. That keeps `useProject()` and `useDatasets()` working for resources-only apps. A dev warning when config and `resources.default` point at different projects would also help until the follow-up lands.

cngonzalez: The deeper fix is in https://github.com/sanity-io/sdk/pull/1041 :) It should be merged first

ryanbonial (dismissed):

ryanbonial (dismissed):

ryanbonial (approved):

PR #1031, opened 2026-07-10
fix(e2e): use correct var and task for dev:e2e

cngonzalez:
### Description
In debugging some failing e2e tests in another branch, I discovered that the `dev:e2e` task wasn't updated with the rest of the recent changes to use `SANITY_APP` variables. This PR updates the command-line variable and uses a dedicated e2e task.

### What to review
Everything make sense?

### Testing
Not really possible to write a test; if you pull down and run `pnpm dev:e2e` you should see that the Kitchensink in staging no longer tries to look for projects in prod.

### DIFF
--- a/apps/kitchensink-react/package.json
+++ b/apps/kitchensink-react/package.json
@@ -8,6 +8,7 @@
     "clean": "rimraf dist",
     "deploy": "sanity deploy",
     "dev": "sanity dev",
+    "dev:e2e": "sanity dev",
     "install": "npm run schema:extract && npm run types:generate",
     "lint": "eslint .",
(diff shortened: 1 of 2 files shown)


apps/kitchensink-react/package.json
```diff
@@ -8,6 +8,7 @@
     "clean": "rimraf dist",
     "deploy": "sanity deploy",
     "dev": "sanity dev",
+    "dev:e2e": "sanity dev",
```
cngonzalez, line 11: additional task needed so we don't pollute the vanilla `dev` namespace in turbo.json with `_E2E_*` type variable allowances

binoy14 (approved):

PR #1034, opened 2026-07-13
feat(core): new fetcher store design with invalidation and optimistic updates

joshuaellis:
## Why

The current fetcher store fetches once per key and keeps that value until the last subscriber unsubscribes. It has no invalidation and no way to reconcile after a mutation, so anything built on it can't refresh stale data or reflect a write without a full reload.

## What it does

Adds a new fetcher store with:

- **Invalidation and staleness** — entries have a stale time and expiry, revalidate in the background when stale, can be invalidated by tag when related data changes, and can poll on an interval.
- **Optimistic updates** — a mutation writes to the cache immediately, reconciles with the server response, and rolls back if the request fails.
- **A single read shape** — reads return `{data, error, isPending, refetch}` (Suspense-first) instead of the raw value.


packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+   * @defaultValue 30_000
+   */
+  staleTime?: number
+  /**
+   * How long an entry survives after its last unsubscribe before removal, in ms.
+   * The clock starts at the last unsubscribe — entries with active subscribers
+   * never expire.
+   * @defaultValue 300_000
+   */
+  expirationTime?: number
```
ryanbonial: nitpick: I would rather see this named something like `gcTime` or `unsubscribeDelay`

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+  id?: string | number
+}
+
+/**
+ * The reactive value of one cache entry, as exposed by
+ * {@link FetcherStore.getState}. React hooks consume this envelope directly.
+ *
+ * @internal
+ */
+export interface FetcherSnapshot<TData> {
```
ryanbonial: Discriminated union 🧐 ?

```suggestion
export type FetcherSnapshot<TData> =
  | {
      status: 'pending'
      data: undefined
      error: undefined
      isFetching: boolean
      dataUpdatedAt: undefined
    }
  | {
      status: 'success'
      data: TData
      error: unknown | undefined
      isFetching: boolean
      dataUpdatedAt: number
    }
  | {
      status: 'error'
      data: undefined
      error: unknown
      isFetching: boolean
      dataUpdatedAt: undefined
    }
```

packages/react/src/hooks/helpers/createFetcherHook.ts
```diff
@@ -0,0 +1,63 @@
...
+   * A background failure (revalidation, poll, or refetch) while cached data still
+   * renders. Initial-load errors throw to the error boundary instead.
+   */
+  error: Error | undefined
+  /**
+   * A fetch for this entry is in flight: background revalidation, polling, a
+   * param-change transition, or a manual `refetch()`. Current data stays on
+   * screen throughout.
+   */
+  isPending: boolean
```
ryanbonial: `isPending` combines React transitions with background fetching. Could we use `isPending` for transitions and isFetching for network activity? Consumers likely need different UI for each.

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+  mutationFn: (instance: SanityInstance) => (input: TInput) => Observable<TResult>
+  /** Optimistic cache writes, applied synchronously before the request */
+  onMutate?: (write: CacheWriter, input: TInput) => void
+  /** Reconcile caches with the server response, avoiding a refetch round-trip */
+  onSuccess?: (write: CacheWriter, result: TResult, input: TInput) => void
+  /**
+   * Cache tags to invalidate once the mutation settles. A static array, or a
+   * function of the result/input.
+   */
+  invalidates?: CacheTag[] | ((result: TResult, input: TInput) => CacheTag[])
```
ryanbonial, line 239: The comment says invalidation runs after success or failure, but this callback requires a `TResult`, which does not exist on failure. Could it receive a settled outcome instead?

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+   */
+  fetcher: (instance: SanityInstance) => (...params: TParams) => Observable<TData>
+  /** Converts params into the cache key that identifies an entry */
+  getKey: (instance: SanityInstance, ...params: TParams) => string
+  /**
+   * Cache tags this entry carries, for cross-store invalidation. A static array,
+   * or a function of the fetched data when tags depend on the response (e.g. one
+   * tag per list item).
+   */
+  tags?: CacheTag[] | ((data: TData, ...params: TParams) => CacheTag[])
```
ryanbonial, line 105: A list needs its `LIST` tag before data arrives and item tags afterward. Could this support both parameter-derived and result-derived tags? Otherwise invalidation cannot match pending entries.

ryanbonial (commented):
A few comments

packages/core/src/store/fetcherStore.ts
joshuaellis: I disagree, I think for a list you would likely use a func to return an array where the first item has type LIST followed by the single items therefore if any of those tags are invalidated the query would run - thoughts?

packages/core/src/store/fetcherStore.ts
joshuaellis: I think the comment may need adjusting in that case. I don't see the use of of triggering in validations if the mutation fails do you?

packages/react/src/hooks/helpers/createFetcherHook.ts
joshuaellis: To me the only transition would be when the data is being fetched no? Or are you thinking of the polling case where you may not necessarily need the UI to react?

I think if you were reacting to user input the mutation would cause a transition so you'd have to listen to the mutations promise but that probably wouldn't actually affect the data hook? 

I might be off on this one - lmk what you think

packages/core/src/store/fetcherStore.ts
joshuaellis: Yum

packages/core/src/store/fetcherStore.ts
joshuaellis: I had gcTime previously, but tried to use the existing language so happy to change it to gcTime!

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+
+import {type SanityInstance} from './createSanityInstance'
+import {type StateSource} from './createStateSourceAction'
+
+/**
+ * Identifies cached data for invalidation. A store entry declares the
+ * tags it carries (via {@link FetcherStoreDefinition.tags}); a mutation declares
+ * the tags it invalidates (via {@link MutationDefinition.invalidates}).
+ *
+ * When invalidating,`{type: 'canvas'}` matches every entry carrying any `canvas` tag;
```
cngonzalez: When might a single store might be carrying one entry for a `canvas` resource, another for a `projectId:dataset` resource, etc.? We often cache things from different resources in _different_ stores, which might mean that the reasoning for this might be a layer above. But there may certainly be endpoints that deliver data useful to a number of different resources that I'm unaware of.

(Or is this distinction more for consumers?)

packages/core/src/store/fetcherStore.ts
cngonzalez: `gcTime` is clear to me too :)

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+   */
+  expirationTime?: number
+  /**
+   * Refetch on this interval (ms) while the entry has at least one subscriber,
+   * independent of staleness. Pauses while `document.hidden`; resumes on
+   * visibility. Off by default.
+   */
+  refetchInterval?: number
+  /**
+   * Consulted once, when an entry is being created and no cached data exists for
```
cngonzalez, line 125: Sorry, could I also get an example for this? I'm assuming it's something like:
1. `ApplicationsStore` exists, has data
2. Consumer calls for data for `applicationX` from `ApplicationStore`
3. `ApplicationStore` doesn't have an entry for `applicationX` yet
4. Its `initialData` function looks like `(applicationId) => applicationsStore.getState()[applicationId]`?

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+  resolveState: (instance: SanityInstance, ...params: TParams) => Promise<TData>
+  /**
+   * Marks the entry stale. If it has active subscribers it refetches immediately;
+   * otherwise it refetches on next subscribe.
+   */
+  invalidate: (instance: SanityInstance, ...params: TParams) => void
+  /** {@link FetcherStore.invalidate} for every entry in this store. */
+  invalidateAll: (instance: SanityInstance) => void
+  /** Imperative fetch, bypassing staleness. Dedupes against an in-flight fetch. */
+  refetch: (instance: SanityInstance, ...params: TParams) => Promise<TData>
```
cngonzalez, line 164: I wonder if we could tighten the model / contract a bit here to ensure that the return from `refetch` comes from the store rather than the return of the fetch? Just to ensure that data remains consistent.

(It may not be possible, just trying to enforce best practices early)

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,218 @@
...
+  data: TResult
+  /** Settles when the refetches of matched active entries complete */
+  invalidated: Promise<void>
+}
+
+/**
+ * Defines a mutation with optimistic cache updates. Lifecycle:
+ *
+ * 1. `onMutate` applies optimistic writes synchronously — the UI updates before
+ *    the request leaves. In-flight refetches for touched entries are ignored on
```
cngonzalez: > In-flight refetches for touched entries are ignored on arrival

What's the mechanism behind this? Will there a per-entry hash or timestamp to represent that an entry has been bumped by an `onMutate`?

cngonzalez (commented):
Mostly just clarifying questions to make sure I understand the model correctly. I've got two overarching assumptions:

1. I feel like much of this is implying that many of the resources using this factory are resources that have two different CRUD endpoints, like `/applications` and `/applications/{application-id}`, and those endpoints will be represented by different stores (i.e., `ApplicationsStore` and `ApplicationStore`). This was my understanding by looking at functions like `initialData` but let me know if I'm off-base here. I got a bit confused by the `LIST` sentinel and thought maybe entries would be co-located, but I don't think that's correct.

(I don't think this model is off-base -- if anything it's kind of like `queryStore` which holds a bunch of things and `projectionStore` which subscribes to some of things)

2. I know that this proposal is making up for things the SDK lacks now -- specifically some reactivity and ability to edit around global resources. Do we envision most of this proposal is for facility around editing, or do you see a need for more of a "we pull data" path for Workbench (as opposed to our current model, which is pretty oriented around "data gets pushed to us")

packages/core/src/store/fetcherStore.ts
joshuaellis: I think it'll be simpler – probably just a version counter, incase two things happen in the same ms for whatever reason.

packages/core/src/store/fetcherStore.ts
joshuaellis: I'm not sure we need to change the interface here, but i'll add a comment to highlight the specific behaviour we're discussing...?

packages/core/src/store/fetcherStore.ts
joshuaellis: yeah the mechanism is essentially what youre saying, but we wont have two stores – here's some pseudo code:

```ts
const applications = defineFetcher({
  name: 'applications',
  fetch: (instance) => () => fetchApplications(instance), // Application[]
  getKey: () => 'LIST',
})

const application = defineFetcher({
  name: 'application',
  fetch: (instance) => (id) => fetchApplication(instance, id), // Application
  getKey: (_instance, id) => id,
  // no `id` entry cached yet → pull it out of the already-cached list
  initialData: (instance, id) => {
    const list = applications.getState(instance).getCurrent()
    const app = list.data?.find((a) => a.id === id)
    return app && list.dataUpdatedAt
      ? {data: app, dataUpdatedAt: list.dataUpdatedAt} // list's fetch time, not "now"
      : undefined
  },
})
```

joshuaellis:
> Mostly just clarifying questions to make sure I understand the model correctly. I've got two overarching assumptions:
> 
> 1. I feel like much of this is implying that many of the resources using this factory are resources that have two different CRUD endpoints, like `/applications` and `/applications/{application-id}`, and those endpoints will be represented by different stores (i.e., `ApplicationsStore` and `ApplicationStore`). This was my understanding by looking at functions like `initialData` but let me know if I'm off-base here. I got a bit confused by the `LIST` sentinel and thought maybe entries would be co-located, but I don't think that's correct.
> 
> (I don't think this model is off-base -- if anything it's kind of like `queryStore` which holds a bunch of things and `projectionStore` which subscribes to some of things)
> 
> 2. I know that this proposal is making up for things the SDK lacks now -- specifically some reactivity and ability to edit around global resources. Do we envision most of this proposal is for facility around editing, or do you see a need for more of a "we pull data" path for Workbench (as opposed to our current model, which is pretty oriented around "data gets pushed to us")

1. We'll have one cache store for all the fetching, like queryStore. I'd think of it like `react-query` to be honest.

2. More the latter, with editing layered on top. Everything(?) reactive in the SDK is document-shaped and push-oriented. Now we're giving global API stuff like (user-applications, projects, etc.) the same reactive read ergonomics documents already have — suspend, cache, stale-while-revalidate, refetch — without a live channel behind them. The optimistic-mutation half is there because once you can pull and cache these resources, editing them needs cache reconciliation to feel instant.

Hope that answers your questions 🤔

packages/core/src/store/fetcherStore.ts
joshuaellis: I think canvas is probably a confusing example, i've changed it to organizations _but_ – it's one store holding all of them (organizations, applications), and type is exactly the "layer above",  it namespaces entries so invalidation hits just the organization ones without a store-per-resource.

I hope i understood your question, i found it a bit hard to read 😅

cngonzalez: Ah -- I think I missed the detail where this is all one large store. I suppose it makes sense since there's no "resource" backing these; I think I was probably envisioning some "artificial" resource for separation or something (a worse decision, do not listen to it)

joshuaellis: It wasn't clear that this was the idea so thanks for raising 🙇🏻‍♀️

packages/core/src/store/fetcherStore.ts
cngonzalez: Was this going to be changed or did I miss a conversation where we decided to keep it?

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,780 @@
...
+      markStale(state, key)
+      if (entry.subscriptions > 0)
+        detach(startFetch(state, def, entry.instance, entry.params, true))
+    }
+  })
+
+  const refetch = bindActionGlobally(
+    cacheStore,
+    ({state, instance}: StoreContext<CacheState>, ...params: TParams): Promise<TData> =>
+      startFetch(state, def, instance, params, true) as Promise<TData>,
```
cngonzalez: Will this (as well as the `baseEntry` call on line 728 in `setData`) create entries with no unsubscription / expiration timers?

I think it's okay that they're making entries with `subscriptions: 0`; it's technically correct because there's technically nothing subscribing to them _yet_. But we only start the expiration process in the `onSubscribe` teardown or an invalidation, which might not happen. Perhaps I am missing something, though?

packages/core/src/store/fetcherStore.ts
```diff
@@ -0,0 +1,780 @@
...
+ * subscriber suspends.
+ *
+ * @internal
+ */
+export interface FetcherDefinition<TParams extends unknown[], TData> {
+  /** Namespaces this fetcher's keys in the shared cache; used for devtools and request-tag derivation */
+  name: string
+  /**
+   * Creates the observable that fetches the data. Transport-agnostic: HTTP
+   * client, comlink node, anything that yields an observable.
```
cngonzalez: Maybe "Comlink node `fetch` function" to be a bit clearer? I noticed from the Claude comment on this PR that it seems to confuse this phrase to mean that a `FetcherDefinition` `fetch` could wrap a streaming / unclosing function which isn't the case. Humans might make that mistake too, I guess.

(Or would make something pretty broken -- anyone trying to make a fetcher based on that would have to do a `first` or a `take` to get things working I think)

cngonzalez (changes requested):
Generally looking good! 2 small nits on naming / comments and one bigger one on staleness that I could have misunderstood.

packages/core/src/store/fetcherStore.ts
joshuaellis: Yeah i think this comment should be better written, it's a bit garbled. Thanks for flagging.

packages/core/src/store/fetcherStore.ts
joshuaellis: 🤔  yes, this has been missed – frustrating.

packages/core/src/store/fetcherStore.ts
joshuaellis: 🎯

joshuaellis:
> Generally looking good! 2 small nits on naming / comments and one bigger one on staleness that I could have misunderstood.

changes made!

cngonzalez (approved):

PR #1035, opened 2026-07-13
fix(core): stop retrying live events connections rejected with a 4xx

binoy14:
### Description

- Bumps `@sanity/client` to `^7.23.1`, which stops the client-internal 1s-forever reconnect loop when a live events SSE connection is rejected with a 4xx (sanity-io/client#1226). Stuck apps in this state were generating ~86K requests/day each.
- With the fixed client, the rejection now surfaces as `ConnectionFailedError` with a `status` — and the query store's flat `retry({delay: 1000})` would recreate the same loop from the SDK side. The query store now treats a 4xx `ConnectionFailedError` as fatal (like `DisconnectError`): live updates stop, queries keep serving, no store-wide error. Status-less connection failures (native EventSource, transient network drops) still retry.

FIXES SDK-1891

### What to review

- `packages/core/src/query/queryStore.ts` — the new `ConnectionFailedError` case in the live events `catchError`.
- The 4xx range check: 408/429 are already retried inside the client and never surface here.

### Testing

- New test: a 401 `ConnectionFailedError` stops live updates without retrying or erroring the store.
- New test: a status-less `ConnectionFailedError` still retries (pins that the fatal check doesn't over-reach).

### Fun gif

Hang up the phone

### DIFF
--- a/packages/core/src/query/queryStore.test.ts
+++ b/packages/core/src/query/queryStore.test.ts
@@ -1,4 +1,10 @@
-import {DisconnectError, type LiveEvent, type SanityClient, type SyncTag} from '@sanity/client'
+import {
+  ConnectionFailedError,
+  DisconnectError,
+  type LiveEvent,
+  type SanityClient,
+  type SyncTag,
+} from '@sanity/client'
 import {delay, filter, firstValueFrom, Observable, of, Subject} from 'rxjs'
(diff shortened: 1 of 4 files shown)


binoy14:

- #1035 ◀

cngonzalez (approved):
LGTM

PR #1039, opened 2026-07-14
feat(react): authenticate AuthBoundary from the workbench OS

joshuaellis:
Inside the workbench, the OS owns the session — there's no login redirect to fall back on. `AuthBoundary` now takes its token from the OS and asks it to reissue on a 401, instead of showing a login screen it should never render.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -62,6 +62,7 @@
     "@sanity/message-protocol": "catalog:",
     "@sanity/sdk": "workspace:*",
     "@sanity/types": "catalog:",
+    "@sanity/workbench": "0.1.0-alpha.24",
     "groq": "catalog:",
     "react-compiler-runtime": "catalog:",
     "react-error-boundary": "catalog:",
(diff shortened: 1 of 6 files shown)


packages/react/package.json
```diff
@@ -62,6 +62,7 @@
     "@sanity/message-protocol": "catalog:",
     "@sanity/sdk": "workspace:*",
     "@sanity/types": "catalog:",
+    "@sanity/workbench": "0.1.0-alpha.24",
```
ryanbonial, line 65: Should this be an optional peer dependency since it is a conditional import?

joshuaellis: Good q – its conditional at the moment, but we need to talk about (in Q3) the DX of delivering workbench APIs and we might end up doing more on it.

IMO i'd keep it here, only because it makes for a lower barrier of entry when people migrate apps etc. but open minded on why not

ryanbonial (approved):

PR #1045, opened 2026-07-16
feat(core)!: add global resource stores

joshuaellis:
## Why

The SDK can only read projects, organizations and datasets, and those stores fetch once and never refresh — stale data can't revalidate and a write can't be reflected without a reload. This implements the resource-store catalogue on the fetcher infrastructure from #1034, giving the SDK read access to the platform's org-level resources with real cache behaviour.

## What it does

- Adds read-only fetchers for organizations, projects, applications, installations, and user-applications (list + detail each), plus an Access API check for the current user's permissions on a resource. All are internal.
- Rebuilds the projects and organizations stores on the new infrastructure: stale-while-revalidate reads, tag-based invalidation, and project detail reads seeded from the already-cached list.

**Breaking:** `getProjectsState`, `resolveProjects`, `getProjectState`, `resolveProject`, `getOrganizationsState`, `resolveOrganizations`, `getOrganizationState` and `resolveOrganization` are no longer exported. The React hooks keep their existing return shapes until the new hook layer lands (SDK-1448).

### DIFF
--- a/packages/core/src/_exports/index.ts
+++ b/packages/core/src/_exports/index.ts
@@ -5,6 +5,8 @@ import {type SanityProject as _SanityProject} from '@sanity/client'
  */
 export type SanityProject = _SanityProject
 
+export {checkPermissions} from '../access/checkPermissions'
+export {type AccessResourceType} from '../access/checkPermissions'
 export type {
   AgentGenerateOptions,
   AgentGenerateResult,
@@ -24,6 +26,25 @@ export {
   agentTransform,
   agentTranslate,
 } from '../agent/agentActions'
+export {application, applications} from '../applications/applications'
+export {
+  type Application,
+  type ApplicationAccess,
+  type ApplicationBase,
+  type ApplicationDeployment,
+  type ApplicationInclude,
+  type ApplicationInterface,
+  type ApplicationsOptions,
+  type ApplicationsResponse,
+  type ApplicationStudioConfig,
+  type ApplicationWorkspace,
+} from '../applications/applications'
+export {userApplication, userApplications} from '../applications/userApplications'
+export {
+  type UserApplication,
+  type UserApplicationDeployment,
+  type UserApplicationsOptions,
+} from '../applications/userApplications'
 export {isStudioConfig} from '../auth/authMode'
 export {AuthStateType} from '../auth/authStateType'
 export {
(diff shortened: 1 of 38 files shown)


joshuaellis:
@claude action both those items you raised

packages/core/src/applications/applications.ts
```diff
@@ -0,0 +1,292 @@
...
+ */
+export interface ApplicationStudioConfig {
+  projectId: string
+  /** Release channel (`next`, `stable`, `latest`) or a pinned semver; null if unset */
+  autoUpdatingVersion: string | null
+}
+
+/**
+ * A deployment of a user application, embedded when `activeDeployment` is
+ * included on the User Applications API. Distinct from the org Applications
```
cngonzalez: Sorry, I thought THIS file was for the "org Applications"? and `userApplications for the "User Applications API"? This belongs in the other file maybe?

packages/core/src/_exports/index.ts
```diff
@@ -24,6 +25,20 @@ export {
   agentTransform,
   agentTranslate,
 } from '../agent/agentActions'
+export {
+  type Application,
```
cngonzalez, line 31: Do we consider the hooks that will use these to be public / things we want the average user to be using? Or should it be moved to `_internal`?

packages/core/src/_exports/_internal.ts
```diff
@@ -6,8 +9,13 @@ export {
   getClientErrorApiType,
   isProjectUserNotFoundClientError,
 } from '../auth/utils'
+export {installation, installations} from '../installations/installations'
+export {organization} from '../organization/organization'
+export {organizations} from '../organizations/organizations'
 export {PREVIEW_PROJECTION} from '../preview/previewConstants'
 export {transformProjectionToPreview} from '../preview/previewProjectionUtils'
+export {project} from '../project/project'
```
cngonzalez: `useProject`, `useProjects` are (I think?) public; are there plans to change them? No strong opinion, just noting it for consistency

packages/core/src/_exports/_internal.ts
```diff
@@ -6,8 +9,13 @@ export {
   getClientErrorApiType,
   isProjectUserNotFoundClientError,
 } from '../auth/utils'
+export {installation, installations} from '../installations/installations'
```
cngonzalez: Slightly confused about functions being `_internal` but type exports being mainline? Thinking --
installations, applications, userApplications, permission checks are internal and projects and orgs being mainline (and all associated exports go along with them?)

cngonzalez (changes requested):
Some nits about comments and exports, otherwise LGTM!

packages/core/src/_exports/_internal.ts
joshuaellis: Yeah I fudged this up a few times. IMO it should all be public & marked as internal in JSDOC - thoughts?

I would assume at some point people will use these, they are just wrappers around our API.

cngonzalez: Yeah, all public is fine -- I'm ok with it especially if they have accompanying React hooks

packages/core/src/applications/applications.ts
joshuaellis: for some reason, that interface was in the wrong file.

packages/core/src/_exports/_internal.ts
```diff
@@ -9,17 +9,6 @@ export {
 export {PREVIEW_PROJECTION} from '../preview/previewConstants'
 export {transformProjectionToPreview} from '../preview/previewProjectionUtils'
 export {getQueryKey, parseQueryKey} from '../query/queryStore' // only used for memoizing in React, not needed for actual functionality
-export {
```
cngonzalez, line 12: Should we leave these here since it's just our internal wiring and not really exposed to users (hopefully)?

cngonzalez (changes requested):
being annoying by asking sdk internals to stay in internals but lmk if there's an issue

packages/core/src/_exports/_internal.ts
joshuaellis: I'm happy to follow your lead on this one, I followed what we do with other store creation which is `public` from core. So lmk the direction 🫡

cngonzalez: I think those predate our `_internal` subentry and should probably be moved too; they're just kept there since we can't do breaking changes. But... now we can 😈

joshuaellis: Great, i'll move those over, and we can review the rest.

cngonzalez (approved):

PR #1046, opened 2026-07-16
feat(core): add operational patch channel for concurrent document edits

ryanbonial:
### Description

Concurrent edits to the same field can silently overwrite or corrupt each other: `useEditDocument` collapses every edit into a whole-field `set`, core re-diffs it into position-addressed patches against a possibly stale snapshot, and rebases re-run that flawed merge. This PR adds an opt-in operational patch channel so collaborative editors (Portable Text first) can exchange real operations instead of snapshots.

Two additive pieces:

- **Read side**: `RemoteDocument` now carries the listener's raw `mutations`, and `applyRemoteDocument` emits a new `remote-patches` document event with the transaction's patch operations and an `origin` (`'local'` for our own transactions observed back through the listener, `'remote'` for other clients).
- **Write side**: `editDocument(doc, patches, {preserveOperations: true})` forwards the given patches verbatim to the server and applies them as-is locally, instead of re-deriving them via `diffValue`. Keyed array operations stay addressable, so concurrent edits interleave instead of last-writer-wins. During a rebase, preserved patches re-apply operationally; if one can no longer apply, the transaction is skipped and surfaced as a `rebase-error` event rather than failing the store.

Nothing changes for existing callers: without the flag, the diff-based path is untouched, and the new event is purely additive.

### What to review

- `packages/core/src/document/reducers.ts`: `remote-patches` emission and origin detection via `unverifiedRevisions`.
- `packages/core/src/document/processActions/edit.ts` and `shared.ts`: the `preserveOperations` branch and `ActionError` wrapping for rebase safety.
- `packages/core/src/document/actions.ts`: the new `editDocument` options parameter.

### Testing

Unit tests for the event emission (origins, sync events, non-patch mutations), verbatim passthrough (draft and liveEdit paths), failure wrapping, and rebase behavior (skip-with-event on conflict, keyed re-anchoring on divergence). An integration test in `documentStore.test.ts` has two store instances concurrently inserting into the same keyed array with `preserveOperations` and asserts both edits land and both instances converge. The two failing tests on CI-green `main` (`liveEvents`, `queryStore`) are unrelated and pre-existing.

### Fun gif

real


packages/core/src/document/processActions/edit.ts
```diff
@@ -71,6 +72,28 @@ export function handleEdit(
     })
   }
 
+  const applyMutations = (
```
cngonzalez: can this be consolidated to the one in `shared.ts` but with an optional param for `documentSetName`?

cngonzalez (dismissed):
LGTM, generally grok the extracting and forwarding patches, tracking own transaction ids, and sending un-transformed patches to the server. One (extremely inconsequential, ignorable) style nit

packages/core/src/document/processActions/edit.ts
ryanbonial: Done!

cngonzalez (approved):

PR #1048, opened 2026-07-17
feat(react)!: return fetch state and refetch from resource hooks

joshuaellis:
## Why

The React resource hooks were rebuilt on the new fetcher store in #1045, but a temporary shim flattened each fetcher snapshot back to bare `data` — so the fetch state the store now tracks (background revalidation, errors) and its `refetch` were unreachable from the UI. This is the hook layer #1045 deferred: it consumes the snapshot directly and drops the shim.

## What it does

- Adds `createFetcherHook`, which builds a Suspense hook over a fetcher using `useSyncExternalStore`, deduping the store's per-commit snapshots so an unchanged entry doesn't re-render, and removes the temporary `mapStateSource` shim.
- `useProject`, `useProjects`, `useOrganization` and `useOrganizations` now return `{data, isFetching, error, refetch}` — they still suspend until the first fetch succeeds, so `data` is always present, while `isFetching` tracks background revalidation and `refetch()` forces a refresh.
- Adds internal hooks over the remaining global resource stores: `useApplications`/`useApplication`, `useUserApplications`/`useUserApplication`, `useInstallations`/`useInstallation`, and `useCheckPermissions`.

**Breaking:** `useProject`, `useProjects`, `useOrganization` and `useOrganizations` return `{data, isFetching, error, refetch}` instead of the value directly. Read `.data` for the previous shape.


packages/react/src/hooks/helpers/createFetcherHook.ts
```diff
@@ -0,0 +1,80 @@
+import {type Fetcher, type FetcherSnapshot} from '@sanity/sdk/_internal'
+import {useRef, useSyncExternalStore} from 'react'
+
+import {useSanityInstance} from '../context/useSanityInstance'
+
+/** Structural equality over the fields that make one snapshot meaningfully different from another. */
+function isSameSnapshot<TData>(a: FetcherSnapshot<TData>, b: FetcherSnapshot<TData>): boolean {
```
cngonzalez: would the `isDeepEqual` from `@sanity/sdk/_internal` help here?

cngonzalez (approved):
generally LGTM. now that we've agreed on consolidating all these to be public we can change the verbiage Claude points out to not call them internal? also one nit to save you like 4 LOC if applicable

packages/react/src/hooks/helpers/createFetcherHook.ts
joshuaellis: yep, lets use what exists today ✅

PR #1049, opened 2026-07-17
feat(react): add mutation hook layer

joshuaellis:
## Why

The resource hooks from #1048 are read-only. The fetcher store gained mutation
support — optimistic write, rollback, tag invalidation — in #1034, but nothing
on the hook layer exposes it, so React can't issue a write or reflect one
without a reload. This is the write-side counterpart to #1048.

## What it does

- Adds `createMutationHook`, which wraps a core `defineMutation` into a hook returning `{mutate, isPending, error, data, reset}` — the write-side twin of `createFetcherHook`.
- Adds internal mutations and hooks over the global application stores: `useUpdateApplication`/`useDeleteApplication`, and `useCreateUserApplication`/`useUpdateUserApplication`/`useDeleteUserApplication`. Each invalidates the matching cache tags, so active `useApplication(s)`/`useUserApplication(s)` entries reconverge on server truth after a write.
- Scoped to browser-usable CRUD; the multipart `createApplication` deploy and the service-token install/uninstall endpoints are left out until a product need calls for them.


packages/react/src/hooks/helpers/createMutationHook.tsx
```diff
@@ -0,0 +1,92 @@
...
+        return mutation(instance, input).then(
+          ({data}) => {
+            if (id === callId.current) setState({status: 'success', data, error: undefined})
+            return data
+          },
+          (error: unknown) => {
+            if (id === callId.current) {
+              setState((prev) => ({status: 'error', data: prev.data, error}))
+            }
+            throw error
```
cngonzalez, line 73: Somewhat agree with Claude -- should we throw here or leave it to consumers to check for error on the result?

cngonzalez (commented):
One question about error handling in hooks, otherwise LGTM

packages/react/src/hooks/helpers/createMutationHook.tsx
joshuaellis: if you couldn't await the mutate function I would agree, but because you can await then imo it should throw an error. It then behaves similarly to `mutateAsync` in `tanstack-query`

cngonzalez: fair!

cngonzalez (approved):

PR #1055, opened 2026-07-17
test(core): add two-client concurrency rig with collision matrix

ryanbonial:
### Description

Adds a two-client concurrency rig for the document store: two real store instances connected to a mock Content Lake that holds submitted transactions until the test releases them, so both clients compute their edits against the same base revision (true concurrency, not just fast sequencing). The mock backend applies released transactions with `processMutations` (the SDK's own content-lake-faithful applier) and emits listener mutation events carrying the actual submitted mutations with `transactionId`/`previousRev`/`resultRev` to every connected client.

This is the safety net for the concurrent-editing fixes stacked on top of it (#1056, #1057, #1058, #1059). It covers 13 collision scenarios (concurrent string inserts at same/different offsets, delete-vs-insert, concurrent increments, different fields, Portable-Text-shaped keyed array edits, `preserveOperations` inserts and sets, format-vs-typing, publish-vs-edit) plus 4 rebase scenarios where a client has an in-flight transaction and a queued one, forcing `applyRemoteDocument` to rebase. Every scenario runs in both apply orders: 34 tests.

Outcomes that lose data with the current pipeline are encoded explicitly (`KNOWN_LOSS`, `KNOWN_DIVERGENCE`) and asserted in their current lossy form, so a fix must flip the corresponding expectation deliberately.

The rig also disproved two assumptions behind the original "adopt @sanity/mutate primitives" plan: `diffValue` already emits `diffMatchPatch` for string edits and keyed `insert`s for keyed arrays, and the rebase in `applyRemoteDocument` is a true 3-way merge against each transaction's original base. Concurrent typing already interleaves correctly through the SDK write path and rebase; the genuinely broken behaviors are the ones fixed by the stacked PRs.

### What to review

- Everything is in one new test file, `packages/core/src/document/documentStore.concurrency.test.ts`. No production code changes.
- Check that the mock Content Lake (`createMockContentLake`) is faithful: held submissions, action-to-mutation conversion, listener event shapes.
- Check that the scenario assertions test what their titles claim, and that the `KNOWN_LOSS`/`KNOWN_DIVERGENCE` mechanism forces later fixes to flip expectations rather than silently changing behavior.

### Testing

This PR is itself a test suite. Run it with:

```
pnpm --filter @sanity/sdk exec vitest run src/document/documentStore.concurrency.test.ts
```

All 34 tests pass, and the full core suite stays green.

### Fun gif

matrix?

### DIFF
--- a/packages/core/src/document/documentStore.concurrency.test.ts
+++ b/packages/core/src/document/documentStore.concurrency.test.ts
@@ -1,0 +1,1031 @@
+/**
+ * Two-client concurrency rig for the document store.
+ *
+ * Two real store instances (separate optimistic pipelines, like two machines)
+ * are connected to a mock Content Lake that:
+ *
+ * - holds submitted transactions until the test releases them, so both clients
+ *   compute their edits against the same base revision (true concurrency)
+ * - applies released transactions with `processMutations` (the SDK's own
+ *   content-lake-faithful applier) against the current server state
+ * - emits listener mutation events carrying the actual submitted mutations
+ *   with `transactionId`/`previousRev`/`resultRev`, to every connected client
+ *
+ * Every scenario runs twice: once with client A's transaction applied first
+ * and once with client B's first. After both transactions settle, both
+ * clients must converge to the server's state. Scenario-specific assertions
+ * then check for data loss (marker survival). Outcomes that are lossy with
+ * the current snapshot-diff rebase are documented in `KNOWN_LOSS` so later
+ * phases (DMP squashing, operational rebase) can flip them deliberately.
+ */
+import {
+  type BaseActionOptions,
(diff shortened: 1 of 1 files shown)


ryanbonial:

- #1055 ◀
    - #1056
        - #1057
            - #1058
                - #1059

cngonzalez (dismissed):
Generally makes sense to me, aside from the rebase scenarios. Since they branch on if they're in the known losses (which they are not) I am not exactly sure what they are testing. Again, based on the urgency of this work I won't block it on that, but that part of the testing harness is a bit opaque to me.

cngonzalez (approved):
thanks for addressing the rebases 🫶

PR #1056, opened 2026-07-17
fix(core): preserve inc/dec operational intent through the snapshot diff

ryanbonial:
### Description

Concurrent `inc`/`dec` operations from different clients no longer collapse to last-write-wins.

The outgoing patches sent to Content Lake are re-derived by diffing the document before and after the user's patches apply. That diff collapses an `inc` into a `set` of the resulting number, so two clients incrementing the same counter concurrently resolve to last-write-wins: the server applies `set count=1` twice instead of `inc 1` twice.

`preserveNumericOperations` (new, in `processActions/shared.ts`) runs after the snapshot diff and swaps diffed `set`s back to the user's operational `inc`/`dec` form, so the server applies the increment against whatever value is current at commit time. A `set` is only swapped when its value equals the base value plus the accumulated user deltas for that exact path, so an `inc` combined with a `set` of the same field still submits the diffed `set` (matching current behavior). Applies to both the Actions API path and the liveEdit mutations path.

**Server-semantics note (verified against gradient):** Content Lake fails the whole transaction with a 400 when `inc`/`dec` targets a missing or non-numeric path, unlike `set` which always succeeds. To avoid turning a concurrent field removal into a batched-transaction revert, the emitted patch also seeds the target with `setIfMissing: {path: baseValue}`. Gradient executes `setIfMissing` before `inc`/`dec` within a single patch (`pkg/documents/mutation/patch.go`), so a concurrently removed field degrades to exactly the value the collapsed `set` would have written. A path into a concurrently removed keyed array item can still fail the transaction (`setIfMissing` cannot create array items); that reverts and reports instead of silently resurrecting the item.

Stacked on #1055.

### What to review

- `preserveNumericOperations` in `packages/core/src/document/processActions/shared.ts`: the delta accumulation, the exact-match guard, the `setIfMissing` seeding, and the split into `inc`/`dec` records.
- Its two call sites: `handleEdit` (`processActions/edit.ts`) and `applySingleDocPatch` (liveEdit/release edit path).
- The flipped expectation in `documentStore.concurrency.test.ts`: `counter-inc-inc` moved out of `KNOWN_LOSS` and now asserts both increments count in both apply orders.

### Testing

- New unit tests in `processActions.test.ts`: plain inc/dec preservation, inc+set of the same field staying a set, keyed-array-item targets, the liveEdit mutation path, and the degradation path (applying the emitted patch to a document whose counter was removed concurrently produces the collapsed-set result via `setIfMissing`, instead of a transaction failure).
- The concurrency rig's `counter-inc-inc` scenario now asserts `count === 2` in both apply orders.
- Server semantics (inc-on-missing fails the transaction; per-patch operation order `setIfMissing` before `inc`/`dec`) verified by reading gradient's `pkg/documents/patching/inc.go` and `pkg/documents/mutation/patch.go`.
- Full core suite, repo type check, and lint are green.

### Fun gif

Counter going up

### DIFF
--- a/packages/core/src/document/documentStore.concurrency.test.ts
+++ b/packages/core/src/document/documentStore.concurrency.test.ts
@@ -392,11 +392,6 @@ const getSpanMarks = (doc: RigDocument | null, blockKey: string): string[] | und
  */
 const KNOWN_LOSS: {scenario: string; firstWriter?: 'A' | 'B'; reason: string}[] = [
   {
-    scenario: 'counter-inc-inc',
-    reason:
-      'user-supplied inc ops are re-derived as set ops by the snapshot diff, so concurrent increments collapse to last-write-wins',
-  },
-  {
     scenario: 'preserved-set-text-vs-typing',
     reason:
       'preserveOperations forwards whole-value sets verbatim, so whichever preserved set lands second overwrites the concurrent one (server-side last-write-wins)',
--- a/packages/core/src/document/processActions.test.ts
+++ b/packages/core/src/document/processActions.test.ts
@@ -4,7 +4,7 @@ import {describe, expect, it} from 'vitest'
 
 import {type Action, type DocumentAction} from './actions'
 import {ActionError, processActions} from './processActions/processActions'
-import {type DocumentSet} from './processMutations'
+import {type DocumentSet, processMutations} from './processMutations'
 
 // Helper: Create a sample document that conforms to SanityDocument.
 const createDoc = (id: string, title: string, rev: string = 'initial'): SanityDocument => ({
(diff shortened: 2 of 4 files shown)


ryanbonial:

- #1055
    - #1056 ◀
        - #1057
            - #1058
                - #1059

cngonzalez (approved):
It probably would be good to test the accumulated `amount` logic at some point (I think on line 145?) -- i think the way to reach it would be several `inc` patches in a single patch.

I suspect it would work fine, so definitely not blocking -- just a note!

cngonzalez (dismissed):

cngonzalez (approved):

PR #1057, opened 2026-07-17
fix(core): converge local to server truth on idle fast-forward

ryanbonial:
### Description

Fixes a divergence where a client could be left viewing stale content indefinitely after its own transaction was confirmed.

When a listener event confirms one of our own transactions with a matching `previousRev`, the store fast-forwards: it advances the `remote` snapshot but keeps the optimistic `local` untouched. The server's result can legitimately differ from our prediction, though. Concretely: client B edits the draft, client A publishes; B's edit lands first, so the server's publish carries B's edit into the published document. A's optimistic prediction was computed from A's own (older) draft. A's echo fast-forwards, nothing ever reconciles `local` with `remote`, and A keeps rendering the stale published content until an unrelated foreign event happens to arrive.

The fix: on a fast-forward, when no other pending local work modifies the document (no applied transaction and no in-flight transaction other than the confirmed one changes it), `local` now converges to the server's document. While pending work that modifies this document exists, the optimistic `local` is kept exactly as before and that work's own echo or rebase recomputes it. The pending-work check is per document and tests for modification (`working[documentId]?._rev !== previousRevs[documentId]`, the same test `transitionAppliedTransactionsToOutgoing` uses), not mere presence: pending edits to other documents must not block convergence here (nothing would ever revisit this document's `local`), and a pending draft edit carries the published document in its working set without changing it, so a presence check would block published-doc convergence whenever a draft edit is pending. A deep-equality guard keeps the previous `local` reference when the content is identical, so subscribers don't re-render for equal values.

The concurrency rig (#1055) caught this as its `publish-vs-edit` divergence.

Stacked on #1056.

### What to review

- The fast-forward branch of `applyRemoteDocument` in `packages/core/src/document/reducers.ts`: the per-document `modifiesDocument`/`nothingElsePending` condition and the `isDeepEqual` reference-preservation guard.
- Whether the "keep optimistic local while other work is pending" behavior matches your mental model of the outgoing/applied pipeline; the non-fast-forward rebase path is unchanged.
- The rig change: the `publish-vs-edit` entry is removed from `KNOWN_DIVERGENCE`, so the scenario now asserts convergence in both apply orders.

### Testing

- The concurrency rig's `publish-vs-edit` scenario flips from asserting the documented divergence to asserting that both clients converge to server truth (both apply orders).
- New `reducers.test.ts` unit tests for the fast-forward convergence: converges when nothing is pending, converges when pending work modifies only other documents, converges when pending work carries this document unmodified (the draft-edit-carrying-published case), and keeps the optimistic local when pending applied or in-flight work modifies this document.
- Existing `reducers.test.ts` fast-forward tests still pass unchanged.
- Full core suite, repo type check, and lint are green.

### Fun gif

truth

### DIFF
--- a/packages/core/src/document/documentStore.concurrency.test.ts
+++ b/packages/core/src/document/documentStore.concurrency.test.ts
@@ -415,18 +415,7 @@ const KNOWN_DIVERGENCE: {
   client: 'A' | 'B'
   documentId: string
   reason: string
-}[] = [
-  {
-    scenario: 'publish-vs-edit',
-    firstWriter: 'B',
-    client: 'A',
-    documentId: PUBLISHED_ID,
-    reason:
-      "the publisher's own echo fast-forwards, keeping the optimistic published " +
-      "prediction as `local` even though the server's publish carried B's newer " +
-      'draft edit; nothing reconciles local with remote afterwards',
-  },
-]
+}[] = []
 
 const isKnownDivergence = (
   scenarioId: string,
--- a/packages/core/src/document/reducers.test.ts
+++ b/packages/core/src/document/reducers.test.ts
@@ -696,6 +696,129 @@ describe('applyRemoteDocument', () => {
     expect(newDocState?.unverifiedRevisions?.['txn10']).toBeUndefined()
   })
 
+  describe('fast-forward convergence of local', () => {
+    const docId = getDraftId(DocumentId('doc1'))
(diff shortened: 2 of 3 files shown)


ryanbonial:

- #1055
    - #1056
        - #1057 ◀
            - #1058
                - #1059

cngonzalez (dismissed):
makes sense to me!

cngonzalez (approved):

PR #1058, opened 2026-07-17
feat(core): harden the shared listener for reconnects and multi-document transactions

ryanbonial:
### Description

Three reliability improvements to how listener events are ingested, aimed at reconnects and multi-document transactions.

1. **Listener resume.** The shared listener now passes `enableResume: true` and subscribes to `welcomeback`/`reset`. A reconnect replays the events missed while offline instead of refetching every subscribed document. When the backend cannot resume, it emits `welcome` or `reset`; both trigger a snapshot refetch (the previous behavior). If the API host does not support resume at all, the client falls back to emitting `welcome` on reconnect, which is exactly the old refetch path.
2. **Dedupe and replay discard.** Resumed events are delivered at least once. A duplicate mutation event would never chain onto the current base revision, so it sat in the sequencing buffer until it overflowed into an `OutOfSyncError` (a full refetch plus backoff). Two layers now prevent that: a TTL dedupe cache keyed by `transactionId#documentId` in the shared listener, and chain-discard in `sortListenerEvents`, which orders buffered events into `previousRev`/`resultRev` chains and discards chains the freshly fetched snapshot already reflects (the same design as `@sanity/mutate`'s `sequentializeListenerEvents`; mutate does not export it from its published entry points, so the equivalent lives here).
3. **Multi-document transaction grouping.** The mutation events of one transaction touching several documents (a publish touches draft + published) are buffered until the transaction is complete, then emitted contiguously. Without this, a rebase can run between the two halves and observe a dataset state that never existed on the server (draft deleted, published not yet updated). While a transaction is incomplete, intervening mutation events from other transactions are held in arrival order and released right after it completes (matching Studio's `getPairListener` approach), because the legacy listener pipeline can interleave concurrent transactions; flushing on interleave would defeat grouping exactly where it is needed. Held events are re-processed on release, so a held multi-document transaction gets grouped too, and per-document `previousRev` sequencing downstream tolerates the reordering. Escape hatches keep a lost event from stalling the stream: everything is released in arrival order when a non-mutation event arrives, when the held queue exceeds 100 events, or after a 30s deadline.

**Backend verification (gradient):** resume is implemented in the streaming listener pipeline (`Last-Event-ID` token replay with a roughly 5 minute retention window; `reset` beyond that, which we refetch on). Replay is no-duplicates by design, so the dedupe here is insurance, plus protection on the legacy pipeline whose gap detector explicitly tolerates duplicates. The server also force-drops every listener connection after roughly 4 hours, so resume turns a guaranteed periodic refetch-everything into a seamless replay. Transaction-event contiguity is guaranteed by the streaming pipeline only; the legacy pipeline can interleave concurrent transactions, which is why the grouping buffer holds intervening events instead of giving up.

Stacked on #1057.

### What to review

- `packages/core/src/document/listenerEventOperators.ts` (new): the TTL cache eviction in `dedupeListenerEvents` and the buffer/flush lifecycle in `groupTransactionEvents` (including timer cleanup on unsubscribe).
- `sharedListener.ts`: the listen options (`enableResume`, event list) and where the two operators sit in the shared pipeline.
- `listen.ts`: `reset` now refetches like `welcome`, `welcomeback` deliberately does not; and the reworked `sortListenerEvents` scan body (`toOrderedChains` + `discardChainTo`). The error classes, buffer size, and deadline semantics are unchanged.

### Testing

- New unit tests in `listenerEventOperators.test.ts` (12 tests): dedupe uniqueness, per-document keys, TTL expiry, max-entry eviction, pass-through of non-mutation events; grouping hold/complete, holding intervening events until the open transaction completes, grouping a held multi-document transaction after release, held-queue overflow, flush before non-mutation events, deadline flush of the transaction and its held events, flush on completion.
- New `sortListenerEvents` tests for replay discard: single replayed event, partially replayed chain, and a multi-event replayed chain.
- `sharedListener.test.ts` updated for the new listen options.
- Full core suite (1068 tests), repo type check, and lint are green.
- Live gate: Snorre's Playwright concurrency harness (13 scenarios against a real dataset) run with this stack and with SDK `main` as baseline produced identical scores, confirming no regression on the wire-level behavior.

### Fun gif

Reconnecting without losing anything

### DIFF
--- a/packages/core/src/document/listen.test.ts
+++ b/packages/core/src/document/listen.test.ts
@@ -205,4 +205,60 @@ describe('sortListenerEvents operator', () => {
     const events = await lastValueFrom(source$.pipe(sortListenerEvents(), toArray()))
     expect(events).toEqual([other])
   })
+
+  it('should discard replayed events already reflected in the base', async () => {
+    // The snapshot is at rev2; a resumed listener replays the event that
+    // produced rev2. It should be discarded instead of clogging the buffer.
+    const sync = createSyncEvent('rev2')
+    const replayed = createMutationEvent({
+      id: 'replayed',
+      previousRev: 'rev1',
+      resultRev: 'rev2',
+    })
+    const source$ = of(sync, replayed)
+    const events = await lastValueFrom(source$.pipe(sortListenerEvents(), toArray()))
+    expect(events).toEqual([sync])
+  })
+
+  it('should discard the replayed part of a chain and apply the rest', async () => {
+    // The snapshot is at rev2; the listener replays rev1->rev2 followed by a
+    // genuinely new rev2->rev3. Only the new event should be emitted.
+    const sync = createSyncEvent('rev2')
+    const replayed = createMutationEvent({
+      id: 'replayed',
+      previousRev: 'rev1',
+      resultRev: 'rev2',
+    })
+    const fresh = createMutationEvent({
+      id: 'fresh',
+      previousRev: 'rev2',
+      resultRev: 'rev3',
+    })
+    const source$ = of(sync, replayed, fresh)
+    const events = await lastValueFrom(source$.pipe(sortListenerEvents(), toArray()))
+    expect(events).toEqual([sync, fresh])
+  })
+
+  it('should discard a multi-event replayed chain that leads up to the base', async () => {
+    const sync = createSyncEvent('rev3')
+    // two replayed events that together lead up to the base revision
+    const replayedA = createMutationEvent({
+      id: 'replayedA',
+      previousRev: 'rev1',
+      resultRev: 'rev2',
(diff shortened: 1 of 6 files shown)


ryanbonial:

- #1055
    - #1056
        - #1057
            - #1058 ◀
                - #1059

cngonzalez (dismissed):
logic generally makes sense to me. i think there's maybe some danger in the grouping logic holding up transactions across many documents (in, for example, a bulk action situation for which we don't really have many tests). I think CLaude's review is pointing that out too. 

Doesn't need to be addressed here, but just noting it as a place for a future optimization (or at least testing)

cngonzalez (approved):

PR #1059, opened 2026-07-17
fix(core): stop fabricating array items for unresolvable keyed set paths

ryanbonial:
### Description

Fixes the client-side patch applier fabricating array items the server never creates, which turned out to be the root cause of the `Unexpected node type` crashes in the concurrent Portable Text editing harness.

Content Lake never creates array items when applying a `set`: a keyed (`[_key=="…"]`) or indexed segment that does not resolve is a silent no-op server-side (verified against gradient's jsonpath matcher, which only auto-creates missing object properties, `pkg/documents/mutation/jsonpath`). The SDK's local applier instead appended a degenerate item, without even the key from the constraint:

```js
set(doc, {'blocks[_key=="b1"].children[_key=="missing"].marks': ['m1']})
// before: children grows a phantom {marks: ['m1']} item (no _key, _type, or text)
// after:  no-op, matching the server
```

Under concurrent editing this was a crash factory: client A formats a span that client B's concurrent split already replaced, the `set` targets a `_key` that no longer exists, and every receiving client's local snapshot grows a phantom span-less child. The server stays valid (it no-ops), but the PT plugin's repair sync faithfully converges the editor to the corrupted local snapshot and the editor render layer throws `Uncaught Error: Unexpected node type` (`use-children.tsx`). This was captured directly with crash forensics in the harness: the malformed `{marks: []}` child and the exact patch that fabricated it.

`set` and `setIfMissing` now skip matches whose array-addressing segments do not resolve. Missing object properties (nested included) are still auto-created, matching the server.

Stacked on #1058.

### What to review

- `resolvesArraySegments` in `packages/core/src/document/patchOperations.ts` and its use in `set`/`setIfMissing`. `unset`, `insert`, `inc`/`dec`, and `diffMatchPatch` already no-op on unresolvable targets.
- The flipped expectation in `patchOperations.test.ts`: the old test asserting that an unmatched key constraint creates the item is replaced by the server-parity no-op, plus new cases (unresolvable keyed property path, out-of-bounds index, missing property on an existing keyed item still created).
- Note this applier backs both optimistic local application and the listener echo reconstruction (`processMutations`), so the fix stops phantom items in all local snapshots.

### Testing

- New/updated unit tests in `patchOperations.test.ts`; full core suite (1080 tests), repo type check, and lint green.
- Harness (Snorre's concurrency rig, local editor build with the #2977 fixes + this SDK build): before the fix, `bold-overlap-range` and `link-overlap-range` crashed in roughly 1 of 3 trials with `Unexpected node type`; after the fix, 0 crashes across 10 focused trials plus a full 13-scenario matrix at 2 trials each. The remaining SILENT-LOSS scores are convergence-timing divergences with identical text on both clients and the server (see PR discussion), not data loss or corruption.

### Fun gif

No more phantoms

### DIFF
--- a/packages/core/src/document/patchOperations.test.ts
+++ b/packages/core/src/document/patchOperations.test.ts
@@ -274,10 +274,51 @@ describe('set', () => {
     expect(output).toEqual({a: 1, nonexistent: {path: 999}})
   })
 
-  it('creates an item from a key constraint if the key is not present', () => {
+  // Content Lake never creates array items when setting: a keyed or indexed
+  // segment that does not resolve is a silent no-op server-side. The applier
+  // used to append a degenerate item (without the key, even), which
+  // fabricated array items locally that would never exist on the server.
+  it('does not create an item from a key constraint if the key is not present', () => {
     const input = {items: [{_key: 'item1'}]}
     const output = set(input, {'items[_key=="item2"]': {_key: 'item2'}})
-    expect(output).toEqual({items: [{_key: 'item1'}, {_key: 'item2'}]})
+    expect(output).toEqual({items: [{_key: 'item1'}]})
+  })
+
+  it('does not create an item when setting a property under an unresolvable key constraint', () => {
+    const input = {
+      blocks: [
+        {
+          _key: 'b1',
+          children: [{_key: 's1', _type: 'span', text: 'hello', marks: []}],
+        },
+      ],
+    }
+    const output = set(input, {'blocks[_key=="b1"].children[_key=="missing"].marks': ['mark1']})
+    expect(output).toEqual(input)
+  })
+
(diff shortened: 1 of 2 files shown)


ryanbonial:

- #1055
    - #1056
        - #1057
            - #1058
                - #1059 ◀

cngonzalez (dismissed):
The scenarios make sense to me -- deeply set an object value at whatever depth, and drop the value if we hit an array segment that refers to something that doesn't exist.

The comments are somewhat difficult to parse at first glance (they've got a lot of wordy LLMese) and it might be nice to provide examples rather than having to contend with phrases like "a string segment may descend into undefined" but obviously not worth stopping good work! More of a mental note to run through our own codebase to make things easier for our devs down the line since people won't have this context later.

cngonzalez (approved):

PR #1065, opened 2026-07-22
fix(core): keep in-flight outgoing visible on foreign rebase

ryanbonial:
## Summary
- When a peer mutation arrives while this client's edits sit only in `outgoing` (`applied` empty), re-apply that submitted transaction onto the new remote for **local EDGE** so optimistic paste/typing is not wiped from `local`.
- Clone `base`/`previous` before `processActions` (mutations mutate document sets in place).
- Skip re-apply when `revision === outgoing.transactionId` so own echoes do not double-apply patches.
- keeps the H1 unit pin for outgoing-empty rebase behavior (paste count and peer block text assertions), and removes debug-only tracing from core reducer code.
## Why
Duo paste under concurrent other-block typing could lose in-flight paste copies: rebase seeded `local` from remote alone and ignored `outgoing`, then plugin repair could persist the loss. Mutator keeps submitted work on EDGE; this matches that for the empty-`applied` window.

## Test plan
- [x] `pnpm exec vitest run src/document/reducers.test.ts src/document/documentStore.concurrency.test.ts` (75 passed)
- [ ] Harness: `duo-typing.mjs paste` + solo/blocks/format regression floor against local SDK build (follow-up; Vite dual-context blocked earlier run)

### DIFF
--- a/packages/core/src/document/reducers.test.ts
+++ b/packages/core/src/document/reducers.test.ts
@@ -854,6 +854,220 @@ describe('applyRemoteDocument', () => {
     expect(newDocState?.local).toEqual(remote.document)
   })
 
+  /**
+   * Paste-loss H1 (duo-typing paste): when pastes sit only in `outgoing` (1s
+   * throttle) and a foreign mutation arrives with `applied` empty, rebase must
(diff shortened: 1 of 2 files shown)


ryanbonial:
Closing in favor of a different editing design

PR #1066, opened 2026-07-23
chore: merge v3 branch

joshuaellis:
### Description

Brings the `v3` line into `main`. v3 reworks how the SDK fetches and mutates Sanity data and reshapes the public API around clearer entrypoints. This is a breaking release.

Highlights:

- New fetcher/resource store architecture with cache invalidation and optimistic updates, replacing the old fetcher store.
- Resource hooks now return fetch state and a `refetch`, plus a dedicated mutation hook layer.
- New `agent`, `comlink`, and `dashboard` entrypoints; dashboard-only hooks (`useOrganizationId`, `useNavigate`, Comlink hooks) are scoped to the dashboard entrypoint.
- `AuthBoundary` authenticates via the workbench OS token.
- Adds application and installation stores and `checkPermissions` / `useCheckPermissions`.

Breaking changes:

- Removes the unimplemented intent system, deprecated utils, and the deprecated user-application stores and hooks.
- Splits utils into subdomain and internal exports.
- Requires a React 19.2 baseline.

A v3 migration guide is included to cover the upgrade path.

### Fun gif

sponsored by baby billy.

walton-goggins-hbo-gif-by-the-righteous-gemstones


packages/react/src/context/WorkbenchTokenRefresh.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return children
+}
+
+/**
+ * Provides workbench OS token refresh. No-op outside the workbench, where the
+ * app uses its normal auth flow.
+ * @public
+ */
+export const WorkbenchTokenRefreshProvider: React.FC<PropsWithChildren> = ({children}) => {
```
cngonzalez: just noting some lingering "workbench" language. even though this has the @public tag it's not actually exported so not a huge deal for this PR

cngonzalez (dismissed):
one nit caught my eye but otherwise 🚀

packages/react/src/context/WorkbenchTokenRefresh.tsx
joshuaellis: will do a quick commit to change this language.

cngonzalez (approved):

PR #1072, opened 2026-07-27
chore(docs): start v3 migration guide in repo

cngonzalez:
### Description
We've historically included a migration guide in the repo (although we only had it last time because it was a very early breaking change). Still, I like having it for consistency and it's good to reference as we add a migration guide for our actual website docs.

I've added a stub here to encompass the changes thus far.

### What to review
Does everything read right?

### Testing
Not required.


joshuaellis (approved):
thank you for doing this 🌟

PR #1077, opened 2026-07-30
fix(core): retain own-transaction memory until echo verification

ryanbonial:
### Description

`cleanupOutgoingTransaction` runs at HTTP ack, before the transaction's listener echo arrives. If the acked transaction was the last thing keeping a document's `DocumentState` alive, the whole state, including `unverifiedRevisions` (the memory used to recognize an echo as our own), got deleted immediately. When the echo eventually arrived it was either silently dropped (no state left to apply it to) or, if something resubscribed in the meantime, seen by a fresh bare state that mislabels it `origin: 'remote'`.

`removeSubscriptionIdFromDocument` no longer deletes a document state that still has a revision awaiting its echo (`isAwaitingOwnEcho`). It's kept alive with an empty subscriber list, so the shared listener (keyed off `documentStates`) keeps listening for it. A new `evictOrphanedDocumentStates` sweep, run after every reducer that can settle an in-flight transaction (ack, echo verification, sync-based pruning, revert), retires the state once nothing is left waiting on it. Retention is bounded by the new `UNVERIFIED_REVISION_RETENTION_TIME` (30s, matching the listener's own chaining deadline) so a transaction whose operations all no-op server-side, and therefore never produces an echo, doesn't pin a document (and its listener) for the life of the store.

### What to review

- `isAwaitingOwnEcho` / `evictOrphanedDocumentStates` in `reducers.ts`, and the four call sites that now sweep: `cleanupOutgoingTransaction`, `revertOutgoingTransaction`, and both return points in `applyRemoteDocument`.
- `revertOutgoingTransaction`'s rebuild step: it now skips rebuilding `local` from `remote` for a subscriber-less document that's still awaiting an unrelated own-transaction's echo, since that transaction isn't represented in `working`.
- `UNVERIFIED_REVISION_RETENTION_TIME` in `documentConstants.ts` and whether 30s is the right value.

### Testing

Unit tests in `reducers.test.ts` cover: memory surviving cleanup so a late echo still labels `'local'` and gets evicted after verifying; a stale wait past the retention time evicting instead; and `recentOwnTransactionIds` alone not extending the wait. An integration test in `documentStore.test.ts` exercises the same ack-then-unsubscribe-then-echo race through the real store and listener wiring. Full core suite (1093 tests), `tsc --noEmit`, `eslint`, and `oxfmt --check` all pass.

### Fun gif
wait


packages/core/src/document/reducers.ts
```diff
@@ -685,22 +709,81 @@ export function removeSubscriptionIdFromDocument(
...
+ * transaction still awaiting its echo.
+ *
+ * This is a sweep rather than a decision made where the last subscriber leaves
+ * because `removeSubscriptionIdFromDocument` defers that eviction while an echo
+ * is in flight, and both halves of the condition can change without a
+ * subscriber going anywhere: the echo can verify, be dropped by a sync, be
+ * discarded by a revert, or simply never arrive. Every reducer that can settle
+ * an in-flight transaction sweeps, so the deferred decision is always re-made.
+ *
+ * Note this can evict the very document whose remote event is being processed,
```
cngonzalez, line 770: the whole document? or the document state? If a subscriber loses all relevant document states, what happens?

cngonzalez (approved):
Approving to unblock anything downstream of this; I can't promise that this is done from a fully-informed POV but I understand there is some urgency here.

If you'd prefer I spend time picking this apart then I can do so; I'm leaving the judgment up to you.

PR #1079, opened 2026-07-30
fix(core): reconnect presence and expire stale sessions

ryanbonial:
### Description

SDK presence stops working permanently after the first dropped socket. `@sanity/bifur-client` has no reconnect logic (`createConnect` errors the observable on `onclose`/`onerror`), and `presenceStore` subscribed with a next-only handler, so a single network blip killed the subscription with an unhandled error and presence never returned for the life of the page. This adds reconnect with exponential backoff, plus the stale-session handling that reconnect implies, and fixes four smaller defects in the same read path.

Two alternatives rejected:

- **Persisting the session id in `sessionStorage`** (what Studio does): a tab that inherits another's session storage, via `window.open` or a same-origin iframe, would read back the same id, and the two live clients would then filter each other's announcements out as their own session. The disconnect on unload plus the expiry sweep solve the same "reload looks like a new person" problem without introducing that collision.
- **Expiring on the sender's `lastActiveAt`**: subject to client clock skew, so a peer whose clock runs slow gets expired immediately and one whose clock runs fast never expires. Expiry uses our own receipt time instead.

No public API change. `packages/core/dist/index.d.ts` is byte-identical to `main`, verified by building both. `types.ts` is +7/-0: two members appended to the `PresenceTransport` tuple, so existing two-element destructuring still type-checks.

### What to review

`packages/core` only, all within `src/presence`. No React changes.

- **`bifurTransport.ts`** — `createConnections` is the load-bearing part. `bifur.heartbeats` is `merge(authedConnection$, heartbeats$, responses$)`, so it emits as soon as the socket is open and authorized rather than waiting up to 30s for the first server heartbeat, and it errors when the socket closes. `distinctUntilChanged` deliberately sits *after* `retry` so one operator instance spans reconnects and each live socket yields exactly one generation.
- **`presenceStore.ts`** — on each new connection generation the session map is cleared and a fresh `rollCall` goes out, since peers may have come or gone while we were away. The expiry sweep is gated on the same generation so a long backoff does not expire everyone while offline.
- **One runtime-observable change:** the placeholder for a user whose profile has not resolved yet is now a structurally valid `SanityUser`. It previously used `as unknown as SanityUser` on an object with no `profile`, so `user.profile.displayName` threw on first render. That access pattern now works; `user.displayName` no longer does, but it only ever worked until the profile resolved a few hundred milliseconds later.
- Canvas user resolution no longer deadlocks: `/canvases/:id` was wrapped in `catchError(() => EMPTY)` and consumed with `first()`, so a failed request meant every canvas participant stayed "Unknown user" forever.
- The `beforeunload` listener no longer leaks one subscription per transport, and now covers `pagehide` too, which is what iOS Safari and back/forward-cache navigations actually fire.

### Testing

8 new unit tests, presence coverage goes from 30 to 38.

I mutation-tested the three that matter rather than trusting them by eye: removing the `retry` operator fails both `connections` tests, removing the reconnect wipe fails the recovery test, and neutralizing the sweep predicate fails the expiry test. All restored afterwards.

The expiry tests assert both directions, that a session announcing every 30s survives well past the TTL and that one which goes quiet is dropped, so the sweep cannot pass by simply deleting everything.

Full suite: 1455 pass, 3 skipped. `pnpm ts:check` clean, `pnpm lint` 0 errors, `knip` clean.

Not covered here: presence has no e2e coverage at all, and every unit test mocks the transport, so nothing exercises a real socket yet. The `unload` tests stub a bare `EventTarget` for `window` rather than pulling `jsdom` into `packages/core`, which runs its tests in Node.

### Fun gif
present

### DIFF
--- a/.github/workflows/fallow.yml
+++ b/.github/workflows/fallow.yml
@@ -32,13 +32,20 @@ jobs:
       - name: Audit changed files
         run: |
           BASE_REF="${{ github.event.pull_request.base.ref || 'main' }}"
-          git fetch origin "$BASE_REF" --depth=1 || true
+          # The explicit refspec matters. `git fetch origin <branch>` writes only
+          # FETCH_HEAD, so `origin/<branch>` never appears for any base other than
+          # the default branch, which checkout already fetched. On a stacked PR the
+          # audit then cannot resolve its base and exits 2, which reads as a lint
+          # failure rather than a missing ref.
+          git fetch origin "+$BASE_REF:refs/remotes/origin/$BASE_REF" --depth=1 || true
           AUDIT_OUTPUT=$(pnpm --silent fallow audit --base "origin/$BASE_REF" --format markdown --quiet) || AUDIT_EXIT=$?
+          # Echoed as well as summarised: a failure is otherwise undiagnosable from
+          # the job log, since the summary is only rendered in the web UI.
           {
             echo "## Fallow audit"
             echo ""
             echo "$AUDIT_OUTPUT"
-          } >> "$GITHUB_STEP_SUMMARY"
+          } | tee -a "$GITHUB_STEP_SUMMARY"
           exit "${AUDIT_EXIT:-0}"
 
       - name: Hotspots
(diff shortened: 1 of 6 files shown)


ryanbonial:

- #1079 ◀
    - #1080
        - #1081

joshuaellis (approved):

PR #1080, opened 2026-07-30
feat(core): report presence at document, field, and cursor level

ryanbonial:
### Description

SDK presence has been read-only: `bifurTransport` implements `presence_announce`, but nothing ever called it, and the store discarded incoming `rollCall` events. So an SDK app could see Studio users while remaining invisible to them, and two SDK apps could never see each other. This adds the write half.

Announcing is opt-in. `localLocations` starts `undefined` and the publish pipeline filters on it, so calling `usePresence` never causes an app to start broadcasting. Reporting `[]` is deliberately distinct from never reporting, and means "present, but not in any document".

The cadence deliberately matches the Studio (`presence-store.ts:152`): a location change, a peer's roll call, or a reconnect restarts `timer(0, 30_000)`, through a 200ms audit window. That is not cosmetic. Studio clients share the NATS room `bifur.presence.<projectId>-<dataset>` and announce every 30s, and #1079 added a 90s expiry sweep, so the TTL has to stay a comfortable multiple of the slowest peer's announce interval. A shorter TTL would expire live Studio users between their announcements.

On `path` typing, one alternative rejected: widening `PresenceLocation.path` from `string[]` to `Path`. That would be correct, since keyed segments already arrive whenever a Studio user focuses an array item or a Portable Text span, but it is a breaking change to a published type and would force a major. Instead the read type is left alone with a doc comment recording that it under-describes reality, and the new `ReportPresenceOptions.path` is typed as a full `Path` so array items and Portable Text spans can be addressed. An internal `WirePresenceLocation` carries the accurate type between store and transport.

Stacked on #1079, so review that first and expect this diff to shrink once it merges.

### What to review

`packages/core` only. No React changes, so no hooks yet: that will come later, along with Portable Text cursor rendering and e2e.

- **`presenceStore.ts`, the publish pipeline.** `merge(localLocations$, rollCallRequests$, connections$)` into `switchMap(() => timer(0, REPORT_MIN_INTERVAL))`. The `filter` on `undefined` after `withLatestFrom` is what makes announcing opt-in. Roll-call echoes of our own session are dropped, since Bifur publishes back to the topic we subscribe to.
- **`sendSafely`, and why every dispatch goes through it.** `switchMap` propagates an inner error to the outer stream, and the heartbeat, roll-call responses, and reconnect re-announce all share that one chain. So a single rejected `presence_announce` would have silenced this client permanently while it carried on reading presence normally. The likeliest trigger is the worst one: a dropped socket errors the in-flight request, so the failure and the reconnect share a root cause, and the announce stream would die just before `connections$` recovered. `sendSafely` catches, logs to the `presence` logger namespace, and continues. No retry, because the 30s heartbeat and the reconnect re-announce already cover recovery.
- **`isEqualLocations`** compares by value and ignores `lastActiveAt`, which is regenerated on every report. Comparing by reference (as the Studio does, where the form rebuilds the array each call) would make every repeated report look like a move and restart the heartbeat.
- **`types.ts`.** `PresenceSelection` is declared structurally rather than imported from `@portabletext/editor`, so `packages/core` stays free of React dependencies per AGENTS.md. It is structurally identical to `EditorSelection`; the assignability assertion belongs in `packages/react`, where the editor becomes an optional peer in PR 3, and the doc comment says so.
- `WirePresenceLocation` is `@internal` and absent from the barrel, since nothing in the public API accepts or returns it.

### Testing

13 new unit tests, presence coverage 38 to 51. Full suite 1468 pass, `ts:check` clean, `lint` 0 errors, `knip` clean.

I mutation-tested every new behaviour rather than trusting a green run: dropping the silence guard, the audit window, the roll-call trigger, the roll-call self-filter, value-based location comparison, the 30s interval, and `sendSafely` on both the announce and unload paths each fail exactly the intended test.

Two of those tests came out of review questions rather than being written up front, and both found real problems:

The announce-failure test asserts that a failed send leaves the heartbeat, roll-call responses, and reconnect re-announce all working. It failed before `sendSafely` existed. The unload-failure test covers `pagehide` firing twice, which happens on a back/forward-cache restore, so a failure on the first must not kill the stream.

One bad test of mine also worth flagging, since the trap is easy to fall into again: my first "collapses a burst" test still passed with `auditTime` removed, because `switchMap` already discards reports made in the same tick, so it was really testing cancellation. It now steps the clock 50ms between reports, which is the case that matters, since a typing user produces focus changes tens of milliseconds apart.

Not covered here: nothing exercises a real socket yet, since every test mocks the transport.

### Fun gif
im here

### DIFF
--- a/packages/core/src/_exports/index.ts
+++ b/packages/core/src/_exports/index.ts
@@ -193,10 +193,13 @@ export {
   type OrganizationsOptions,
   resolveOrganizations,
 } from '../organizations/organizations'
-export {getPresence} from '../presence/presenceStore'
+export {getPresence, reportPresence} from '../presence/presenceStore'
 export type {
   DisconnectEvent,
   PresenceLocation,
+  PresenceSelection,
+  PresenceSelectionPoint,
+  ReportPresenceOptions,
   RollCallEvent,
   StateEvent,
   TransportEvent,
--- a/packages/core/src/presence/presenceStore.test.ts
+++ b/packages/core/src/presence/presenceStore.test.ts
@@ -826,6 +826,31 @@ describe('presenceStore', () => {
       }
     })
 
+    it('keeps announcing when a trigger source fails', async () => {
+      vi.useFakeTimers()
+      try {
+        getPresence(instance)
+        reportPresence(instance, {locations: [{documentId: 'doc-1'}]})
+        await flush()
+        expect(stateCalls()).toHaveLength(1)
+
+        // The three announce triggers are combined with `merge`, which propagates
+        // an error from any one of them. The inbound stream feeds the roll-call
+        // trigger, so a failure there would otherwise take the heartbeat and the
+        // reconnect re-announce down with it.
+        mockIncomingEvents.error(new Error('inbound stream failed'))
+
+        await vi.advanceTimersByTimeAsync(30_000)
+        expect(stateCalls()).toHaveLength(2)
+
+        mockConnections.next(2)
+        await flush()
+        expect(stateCalls()).toHaveLength(3)
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+
     it('re-announces after a reconnect', async () => {
       vi.useFakeTimers()
       try {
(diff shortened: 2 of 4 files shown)


ryanbonial:

- #1079
    - #1080 ◀
        - #1081

joshuaellis (dismissed):

joshuaellis (approved):

PR #1081, opened 2026-07-30
feat(react): add presence reporting and per-document presence hooks

ryanbonial:
### Description

Completes the presence work: #1079 fixed the read path, #1080 added the core write path, and this makes both usable from React.

https://github.com/user-attachments/assets/be600647-9ab8-4d31-8b41-34a7b06da623

`useReportPresence` is the opt-in write site, and `usePresenceForDocument` is a per-document read flattened to one entry per participant per location, so it can be rendered straight against a field. `usePresence` still returns every participant in the whole project and dataset and leaves the filtering to you.

Two deliberate departures from the plan:

- **`usePresenceCursors` is not here.** Returning `RangeDecoration[]` means naming a type from `@portabletext/editor`, and since `@portabletext/plugin-sdk-value` already declares `@sanity/sdk-react` as a peer, adding the reverse peer would make the two mutually peer-dependent and tie the SDK's release cadence to the editor's major version. The Portable Text cursor mapping belongs in a `@portabletext/plugin-sdk-presence` where the editor is already a peer. No new dependency in this PR.
- **Document filtering lives in core, not the hook.** Version matching needs `@sanity/id-utils`, which core already depends on and `packages/react` does not. Putting `getDocumentPresence` in core keeps the hook a thin binding, matches the core-owns-state pattern, and avoids a new dependency.

`DocumentPresence.path` is a proper `Path`, which is the payoff from #1080's decision to leave the published `PresenceLocation.path` as `string[]`. New type, so consumers get honest keyed segments without a breaking change to the old one.

Stacked on #1080 which is stacked on #1079.

### What to review

- **`packages/core/src/presence/presenceStore.ts`, the snapshot cache.** `createStateSourceAction` runs its selector uncached on every `getCurrent()`, and `getCurrent` is what React calls as its `useSyncExternalStore` snapshot, so a selector building a fresh array sends React into an infinite render loop. `usePresence` avoids this with `reselect`, but a single memo slot will not do here: the query is parameterised per document, so callers watching different documents would evict each other. Cached in a `WeakMap` keyed on the state object, then per query.
- **User resolution moved into the store's `initialize()`.** It used to hang off `getPresence`'s `onSubscribe`, so reading presence any other way resolved no profiles and every participant showed as "Unknown user". It is driven by which user ids are in state, which is store-level, and this also stops two selectors duplicating the same profile requests.
- **`useReportPresence` throttling.** 1000ms for field focus, 250ms when a `selection` is present, because a caret trailing a full second reads as broken rather than as latency. The reported location is rebuilt from its serialized key so the effect can depend on it honestly rather than suppressing `exhaustive-deps`, and nothing is written to a ref during render.
- **`packages/core/src/presence/paths.ts`.** A `Path` prefix comparison, because the exported path helpers come from `@sanity/json-match` and model paths as JSONMatch expressions, not `(string | number | {_key})[]`. Compares keyed segments by `_key`, which is what actually arrives.
- **The kitchensink route was rewritten rather than extended.** It was three `JSON.stringify` dumps against hardcoded project and canvas ids whose implicit workflow was "open a Studio in another tab", because the SDK could not announce. It now demonstrates presence on its own, defaults to a real document so there is no id to look up, and links to the same document in a Studio.

### Testing

19 new unit tests, presence coverage 53 to 72. Full suite 1493 pass, `ts:check` clean, `lint` 0 errors, `knip` clean.

Every new behaviour was mutation-tested rather than trusted on a green run: removing the snapshot cache, the published-id normalisation, the keyed-segment comparison, the path filter, and the profile write each fail exactly the intended test. Two of those tests exist only because the mutation exposed that nothing covered them.

One finding changed the e2e design. Presence stores are keyed by project and dataset and held in a module-level registry, so two `SanityInstance`s in the same JavaScript realm share one store, one socket, and one session id. I verified this rather than assumed it: two instances on the same resource create exactly one transport. A two-pane harness would therefore have been a single participant and the assertions would have passed while proving nothing, so `presence.spec.ts` drives two browser contexts, which are separate realms and so genuinely separate sessions over a real Bifur room.

Worth knowing: `apps/kitchensink-react/src/routes/PortableTextRoute.tsx` has a comment claiming its two panes "round-trip through the Content Lake listener instead of sharing a document store". By the same registry logic that looks untrue, so that route's concurrency test may be weaker than it claims. Untested and out of scope here.

### Fun gif
im in

### DIFF
--- a/packages/react/src/hooks/presence/useReportPresence.ts
+++ b/packages/react/src/hooks/presence/useReportPresence.ts
@@ -1,5 +1,4 @@
 import {
-  getEditingDocumentId,
   isMediaLibraryResource,
   type PresenceSelection,
   reportPresence,
@@ -81,12 +80,6 @@ export interface UseReportPresenceOptions extends DocumentHandle {
 export function useReportPresence(options: UseReportPresenceOptions): void {
   const {path, selection, throttleMs, ...handle} = options
 
-  // Resolved to the specific document the user is editing under this perspective,
-  // because that is what other clients compare against. The Studio's field
-  // indicators match its form's id exactly, so reporting the published id would
-  // appear at document level and never light up a field.
-  const documentId = getEditingDocumentId(options)
-
   const normalizedOptions = useNormalizedResourceOptions(handle)
   if (normalizedOptions.resource && isMediaLibraryResource(normalizedOptions.resource)) {
     throw new Error(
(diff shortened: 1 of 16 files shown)


ryanbonial:

- #1079
    - #1080
        - #1081 ◀

packages/react/src/hooks/presence/useReportPresence.ts
```diff
@@ -0,0 +1,170 @@
...
+      // Trailing edge, so the position the user settled on is the one announced.
+      clearTimeout(pending.current)
+      pending.current = setTimeout(send, interval - elapsed)
+    }
+
+    return () => clearTimeout(pending.current)
+  }, [location, interval, sanityInstance, resource])
+
+  // Kept separate from the throttled effect so it runs on unmount only, rather
+  // than every time the reported location changes.
```
joshuaellis, line 164: nit: doesn't necessary run on `unmount` when you add those references below, but probably still fine?

joshuaellis (dismissed):
nice.

joshuaellis (approved):

PR #1083, opened 2026-08-05
fix(react): allow `null` as fallback for `ResourceProvider`

gu-stav:
### Description

`ResourceProvider` resolved its `fallback` with `??`, so `fallback={null}` — the idiomatic "render nothing" — landed on the built-in `DEFAULT_FALLBACK` and rendered `Warning: No fallback provided. Please supply a fallback prop…` in place of the suspended tree. 

I think it should be able to receive `null`, to render nothing, which is aligned with how `Suspense` works in React - otherwise I'd have to pass in a Fragment.

### Testing

Unit test: a suspending child under `fallback={null}` leaves the DOM empty.

### DIFF
--- a/packages/react/src/context/ResourceProvider.test.tsx
+++ b/packages/react/src/context/ResourceProvider.test.tsx
@@ -59,6 +59,27 @@ describe('ResourceProvider', () => {
     consoleSpy.mockRestore()
   })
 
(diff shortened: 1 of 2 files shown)


joshuaellis (approved):

cngonzalez (approved):
makes sense to me :)

PR #1085, opened 2026-08-05
feat(core): add commenting store and write actions

ryanbonial:
### Description

The SDK has no comments API, so an app built on it cannot read or write the comments people leave in Studio. This adds the framework-agnostic core layer: addon dataset discovery and provisioning, a realtime comments store, read selectors for flat lists and threads, and the five write actions (create, reply, update, set status, remove).

The written shape is an interop contract, which is the main reason this is worth careful review. Comments live in a per-dataset `comments` addon dataset, and Studio filters on `target.document._ref` and groups on `target.path.field`. If the SDK writes either differently, its comments silently stop appearing in Studio without anything failing.

Alternatives considered and rejected:

- Port Studio's `useCommentsStore` as-is. It is React-bound, and it loses mutations that land while its snapshot fetch is in flight because the older snapshot overwrites them.
- Keep comments in the app's own dataset. Simpler, but Studio would never see them.
- Fetch on demand instead of holding a listener. Drops realtime, which is most of the point of collaborative comments.

Core only. React hooks and the Portable Text inline-comment plugin follow as separate PRs in this stack.

### What to review

Scope is `packages/core` alone: everything new lives under `src/comments/`, plus additive exports in `src/_exports/index.ts`. No dependency changes.

In rough order of how expensive a mistake would be:

- **Interop.** `buildCommentPayload` in `commentActions.ts` and `toCommentFieldPath` in `commentFieldPath.ts`. Both have golden tests pinning the exact strings Studio writes, including the quoting Studio applies to field names GROQ reads as values (`true`, `false`, `null`). If you know Studio's comments code, this is the highest-value read.
- **Optimistic writes.** `commentActions.ts` with `reducers.ts`. A comment appears before the round trip and survives a snapshot landing mid-flight; a failed create stays on screen carrying its error so an app can retry with the same id; edits, status changes, and deletes roll back if the server rejects them. Status changes and deletes each go out as a single mutation so a thread and its replies cannot end up half-updated.
- **Listener behavior.** `observeComments.ts` buffers mutations that arrive while the snapshot fetch is running, rather than letting the older snapshot overwrite them. `commentsStore.ts` keeps following the addon client after a listener error, so a token refresh or reconnect recovers instead of leaving the document permanently unable to load.
- **Loading semantics.** `addonDatasetStore.ts` distinguishes "discovery still running" from "this project has no comments dataset". Collapsing the two would make a reader resolve an empty list on a project that does in fact have comments.

### Testing

Unit tests sit next to each source file. 112 tests across the comments module; the full core suite is green at 1,257.

Golden tests pin the written document shape and the field path serialization to what Studio produces, so drift fails loudly rather than quietly making comments invisible.

The races each have a regression test, written failing first and then fixed: a snapshot arriving against an in-flight create, an echo from a superseded transaction, listener recovery after an error, a read issued while discovery is still in flight, concurrent callers sharing one addon dataset setup call, and rollback on a failed edit, status change, or delete.

Not automated here: provisioning a real addon dataset against Content Lake, and confirming an SDK-written comment renders in Studio. Both need a live project, so they are worth a manual pass before this reaches users.

### Fun gif
comments

### DIFF
--- a/packages/core/src/_exports/index.ts
+++ b/packages/core/src/_exports/index.ts
@@ -91,15 +91,11 @@ export {
   resolveCommentThreads,
 } from '../comments/commentsStore'
 export {
-  type CommentContext,
-  type CommentDocument,
+  type Comment,
   type CommentLocalState,
   type CommentMessage,
-  type CommentPath,
-  type CommentPostPayload,
-  type CommentReactionItem,
+  type CommentReaction,
   type CommentStatus,
-  type CommentTarget,
   type CommentTextSelection,
   type CommentTextSelectionItem,
   type CommentThread,
--- a/packages/core/src/_exports/index.ts
+++ b/packages/core/src/_exports/index.ts
@@ -81,7 +81,6 @@ export {
   updateComment,
   type UpdateCommentOptions,
 } from '../comments/commentActions'
-export {toCommentFieldPath} from '../comments/commentFieldPath'
 export {
   type CommentsOptions,
   getCommentsState,
--- a/packages/core/src/_exports/index.ts
+++ b/packages/core/src/_exports/index.ts
@@ -69,6 +69,41 @@ export {
   type RequestNewTokenMessage,
   type WindowMessage,
 } from '../comlink/types'
+export {
+  createComment,
+  type CreateCommentOptions,
+  removeComment,
+  type RemoveCommentOptions,
+  replyToComment,
+  type ReplyToCommentOptions,
+  setCommentStatus,
+  type SetCommentStatusOptions,
+  updateComment,
+  type UpdateCommentOptions,
+} from '../comments/commentActions'
+export {toCommentFieldPath} from '../comments/commentFieldPath'
+export {
+  type CommentsOptions,
+  getCommentsState,
+  getCommentThreadsState,
+  resolveComments,
+  type ResolveCommentsOptions,
+  resolveCommentThreads,
+} from '../comments/commentsStore'
+export {
+  type CommentContext,
+  type CommentDocument,
+  type CommentLocalState,
+  type CommentMessage,
+  type CommentPath,
+  type CommentPostPayload,
+  type CommentReactionItem,
+  type CommentStatus,
+  type CommentTarget,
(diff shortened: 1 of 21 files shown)


packages/core/src/comments/addonDatasetStore.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+function discover(
+  instance: SanityInstance,
+  resource: DatasetResource,
+): Observable<string | undefined> {
+  return getClientState(instance, {
+    apiVersion: ADDON_DATASET_API_VERSION,
+    projectId: resource.projectId,
+    useProjectHostname: true,
```
cngonzalez, line 83: Do we need `projectId` and `useProjectHostname` here if we're calling `/projects/${projectId}/datasets?datasetProfile=${ADDON_DATASET_PROFILE}&addonFor=${dataset}`? It seems like the project is getting declared twice. 

(https://api.sanity.io/v2025-02-09/projects/ppsg7ml5/datasets?datasetProfile=comments&addonFor=test works fine, by my testing. I don't have a preference for global or not global URL at this point, just noting the redundancy)

packages/core/src/_exports/index.ts
```diff
@@ -69,6 +69,37 @@ export {
...
+  removeComment,
+  type RemoveCommentOptions,
+  replyToComment,
+  type ReplyToCommentOptions,
+  setCommentStatus,
+  type SetCommentStatusOptions,
+  updateComment,
+  type UpdateCommentOptions,
+} from '../comments/commentActions'
+export {toCommentFieldPath} from '../comments/commentFieldPath'
```
cngonzalez: Is this something that is public-facing or is it more of an internal util (i.e., should it be in our `internal` subexport file instead?)

cngonzalez (dismissed):
A few nits, not blocking

packages/core/src/_exports/index.ts
ryanbonial: Fixed. Thanks!

cngonzalez (approved):

PR #1087, opened 2026-08-05
feat(react): add comment hooks

ryanbonial:
### Description
This adds the three hooks for comments. Both read hooks follow `useQuery` rather than the simpler `createStateSourceHook`. Comments are subscription-backed, and `createStateSourceHook` builds a fresh state source every render, which would re-register a subscriber each time. The deferred-key memo avoids that and gives `isPending` for free.

`useCommentActions` resolves the resource when an action is called rather than when the hook runs, matching `useApplyDocumentActions`, so one set of callbacks serves an app spanning several resources.

Stacked on #1085. Base is `comments-core`, not `main`. The Portable Text inline-comment plugin follows in the editor repo once this publishes.

### What to review

Two packages plus the kitchensink app. Nothing outside `src/hooks/comments/`, `src/comments/commentsStore.ts`, the two export manifests, and the new route. No dependency changes.

In rough order of how expensive a mistake would be:

- **`useCommentList.ts`.** The only non-obvious file. It holds a deferred key so a document or filter switch runs in a transition instead of suspending a second time, aborts the previous read so its listener drops when nothing else is reading, and suppresses `react-hooks/refs` on the throw. The suppression is deliberate and matches `useQuery`: React runs no effects for a render that suspends, so the captured signal cannot be swapped underneath it.
- **`getCommentsOptionsKey` in `commentsStore.ts`.** A field missing from this key means a reader silently keeps the list it had while the caller believes it asked for a different one. The `satisfies` intersection makes adding a field to `CommentsOptions` a compile error here unless it is keyed or explicitly named irrelevant.
- **`useCommentActions.ts`.** Resource resolution happens per call. Worth checking the tiers behave as you expect when a call names a resource that context does not have.
- **`CommentsRoute.tsx`.** Read last. Longest file, least surprising.

### Testing

16 hook tests alongside the hooks, 3 more on the core key helpers. The comments module is at 115 core tests and 16 React; the full suite is green at 1,649.

Covered: suspending until the first snapshot, keeping the previous list on screen while a different document loads rather than falling back, perspective inherited from context (dropping it would read and write the wrong release with nothing to show for it), named-resource resolution, callback identity across renders, and per-call resource override.

### Fun gif
comment

### DIFF
--- a/apps/kitchensink-react/src/Comments/CommentsRoute.tsx
+++ b/apps/kitchensink-react/src/Comments/CommentsRoute.tsx
@@ -1,0 +1,546 @@
+import {
+  type CommentDocument,
+  type CommentMessage,
+  type CommentStatus,
+  type CommentThread,
+  useCommentActions,
+  useCommentThreads,
+  useDocumentProjection,
+  useDocuments,
+  useResource,
+} from '@sanity/sdk-react'
+import {
+  Badge,
+  Box,
+  Button,
+  Card,
+  Code,
+  Flex,
+  Inline,
+  Select,
+  Stack,
+  Text,
+  TextInput,
+} from '@sanity/ui'
+import {type JSX, useState} from 'react'
+import {useSearchParams} from 'react-router'
+
+import {PageLayout} from '../components/PageLayout'
+
+const DOCUMENT_TYPE = 'author'
+
+/**
+ * The Studio serving the kitchensink's default resource, `ppsg7ml5` / `test`.
+ * Override with `?studio=<baseUrl>` when pointing at another one.
+ */
+const DEFAULT_STUDIO_BASE_URL = 'https://test-studio.sanity.build/test'
+
+/**
+ * Fields to hang threads off.
+ *
+ * Both exist on the `author` schema. That matters for the interop check: a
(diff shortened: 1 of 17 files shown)


cngonzalez (dismissed):
LGTM! semantically aligned with other sdk react utils :feelsgood:

apps/kitchensink-react/src/Comments/CommentsRoute.tsx
```diff
@@ -0,0 +1,567 @@
...
+
+const OPEN_DISPLAY = {
+  tone: 'default',
+  badge: 'primary',
+  action: 'Resolve',
+  nextStatus: 'resolved',
+} as const
+
+/** Wraps plain text as the Portable Text the Studio stores. */
+function toMessage(text: string): CommentMessage {
```
cngonzalez, line 69: Out of curiosity, do the Studio or other consumers generally expect Portable Text? If this is the case, should we add a param or default to do this on the hook that the user can override (if they want to add some custom PT styles / objects / et al.)

cngonzalez (approved):
One comment about the conversion to / from PTE format but can def be done in follow-up PR if it's the right direction

PR #1096, opened 2026-08-11
chore(deps): bump @sanity/pkg-utils to v12 and migrate package configs

ryanbonial:
### Description

Bumps `@sanity/pkg-utils` from v10 to v12.1.0. v12 swaps the rollup/rolldown stack for tsdown and removes several config options, so the shared and package-level `package.config.ts` files needed a migration before builds would run again.

Migrated `extract` → `tsdoc`, `rollup.plugins` → `plugins`, and the React Compiler options to the top-level `reactCompiler` shape. Dropped `dts: 'rolldown'` (now the default). The first successful `pkg build` also rewrote package exports maps: redundant `import` conditions removed, `publishConfig.exports` added without the `source` condition, and `inlinedDependencies` recorded on `@sanity/sdk-react`.

### What to review

- Catalog bump in `pnpm-workspace.yaml` and lockfile.
- Shared config: `packages/@repo/package.config/src/package.config.ts` (`extract` → `tsdoc`).
- Package configs for `@sanity/sdk` and `@sanity/sdk-react` (plugins / reactCompiler / removed `dts`).
- Auto-updated `package.json` exports and `publishConfig.exports` on core, react, and e2e. Confirm the published map still looks right for consumers.

### Testing

- `pnpm --filter @sanity/sdk --filter @sanity/sdk-react --filter @repo/e2e build` succeeds.
- `pnpm fallow audit` clean for the changed files.
- No new automated tests; this is build tooling only.

### Fun gif

upgrade montage

### DIFF
--- a/packages/core/package.config.ts
+++ b/packages/core/package.config.ts
@@ -7,17 +7,14 @@ const enableVisualizer = process.env['VISUALIZER'] === 'true'
 export default defineConfig({
   ...basePackageConfig,
   tsconfig: 'tsconfig.dist.json',
-  rollup: {
-    plugins: [
-      ...(enableVisualizer
-        ? [
-            visualizer({
-              filename: './stats/index.html',
-              open: false,
-            }),
-          ]
-        : []),
-    ],
-  },
-  dts: 'rolldown',
+  plugins: [
+    ...(enableVisualizer
+      ? [
+          visualizer({
+            filename: './stats/index.html',
+            open: false,
+          }),
+        ]
+      : []),
+  ],
 })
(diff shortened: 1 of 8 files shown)


packages/core/package.config.ts
stipsan, line 1: I have not used `rollup-plugin-visualizer` with rolldown, so I'm not sure which option is best here. Just wanted to flag that there is an experimental built in plugin in rolldown that could be used: [https://rolldown.rs/builtin-plugins/bundle-analyzer](<https://rolldown.rs/builtin-plugins/bundle-analyzer>)<br>Which has the interesting benefit of setting `format: 'md'` so it's ideal for an LLM to look at

stipsan (approved):

mttdnt (approved):

PR #1114, opened 2026-08-25
feat(react)!: move dashboard-only Comlink hooks to dashboard entrypoint

gu-stav:
### Description

`useNavigateToStudioDocument`, `useWindowTitle`, `useAgentResourceContext` are dashboard-specific comlink hooks and are now exported from the dashboard entrypoint.

I left `useWindowConnection`, and `useFrameConnection` and other low-level types/ utilities at the root, because they are not tied to dashboard. `@sanity/sdk` has a separate `comlink` entrypoint and I am guessing that is useful in studio/ presentation?


joshuaellis (approved):

cngonzalez (commented):
Should we move the relevant types too, as Claude suggests?

ryanbonial (approved):
The move itself looks good, but let's make sure we don't lose the typedoc generated documentation for these hooks
https://github.com/sanity-io/sdk/blob/feat/react-dashboard-comlink-hooks/packages/react/typedoc.json#L3

gu-stav:
Good catch, @cngonzalez and @ryanbonial. Moved the two dashboard-specific comlink types too and added the dashboard entrypoint for typedoc.

joshuaellis (approved):

PR #1119, opened 2026-08-26
feat!: v3

joshuaellis:
### Description

Brings the `v3` line into `main`. v3 reworks how the SDK fetches and mutates Sanity data and reshapes the public API around clearer entrypoints. This is a breaking release.

Highlights:

- New fetcher/resource store architecture with cache invalidation and optimistic updates, replacing the old fetcher store.
- Resource hooks now return fetch state and a `refetch`, plus a dedicated mutation hook layer.
- New `agent`, `comlink`, and `dashboard` entrypoints; dashboard-only hooks (`useOrganizationId`, `useNavigate`, Comlink hooks) are scoped to the dashboard entrypoint.
- `AuthBoundary` authenticates via the workbench OS token.
- Adds application and installation stores and `checkPermissions` / `useCheckPermissions`.

Breaking changes:

- Removes the unimplemented intent system, deprecated utils, and the deprecated user-application stores and hooks.
- Splits utils into subdomain and internal exports.
- Requires a React 19.2 baseline.

A v3 migration guide is included to cover the upgrade path.

### Fun gif

sponsored by baby billy.

walton-goggins-hbo-gif-by-the-righteous-gemstones

PR #1121, opened 2026-08-26
fix: fix subscription leak in fetcher store

cngonzalez:
### Description
Claude [caught a bug](https://github.com/sanity-io/sdk/pull/1119#issuecomment-5428402627) in the fetcher store while we were about to merge v3. The subscription added when a fetch began was never able to be cleaned up. Although the `client.observable.request` completes, `getClientState.observable` does not (because of the StateSourceAction logic).

This means any new request would see that old subscription, see it in progress, and return its promise. You can recreate this on the `v3` branch by doing the following:
1. Open your network tab.
2. Go to the `projects` page in Kitchensink (running dev locally)
3. You should see fresh `projects` API calls being made.
4. Navigate somewhere else and wait for 30 seconds (the fetcher store `staleTime`)
5. Go back to the `projects` page
6. No new API calls 😢 

This also left the subscription hanging around, which is also not so good. Adding a `take(1)` here lets the `complete` and teardown in the fetcher store happen.

### What to review
Is `take(1)` the right move? Any side effects? Do we need to do a share or anything?

### Testing
A regression test was added. You can follow the same steps above on THIS branch to see the fix in action.

### DIFF
--- a/packages/core/src/store/fetcherStore.test.ts
+++ b/packages/core/src/store/fetcherStore.test.ts
@@ -1,4 +1,4 @@
-import {NEVER, of, startWith, Subject, throwError} from 'rxjs'
+import {concatWith, NEVER, of, startWith, Subject, throwError} from 'rxjs'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
 
 import {createSanityInstance, type SanityInstance} from './createSanityInstance'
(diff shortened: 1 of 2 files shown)


joshuaellis (approved):
Happy with the regression test you've added.

PR #1125, opened 2026-08-28
feat(react): add typed dashboard URL builder

gu-stav:
### Description

Port the URL builder from the `@sanity/workbench` package, with two differences:

### 1. The builder is now extensible

The internal singleton builders are now built based on the new `UrlBuilder` class, which is also exported, so that developers can build their own grammar based builders.

I believe this is helpful, because there are multiple new singleton applications, that are not available to everyone or experiments and therefore shouldn't show up in the SDK. We will likely need the ability to extend its surfaces though so we build up URLs e.g. in content agent to the an experimental singleton app.

```ts
// persona/urlFor.ts
import {UrlBuilder, urlFor} from '@sanity/sdk-react/dashboard'

export class PersonaUrlBuilder extends UrlBuilder {
  static readonly namespace = 'persona'

  person(personId: string): this {
    return this.append('people', personId)
  }
}

export const personaUrlFor = urlFor.extend({
  persona: PersonaUrlBuilder,
})
```

The Persona package can re-export its extended builder:

```ts
// persona/index.ts
export {PersonaUrlBuilder, personaUrlFor as urlFor} from './urlFor'
```

### 2. Consumers can choose relative or absolute output

Previously it was based on the window origin, which may be surprising. With an origin being passed to `url`, it now returns an absolute URL.

```ts
person.url()
// '/persona/people/person%2F1'

person.url({origin})
// 'https://dashboard.sanity.io/persona/people/person%2F1'

person.toURL({origin})
// URL('https://dashboard.sanity.io/persona/people/person%2F1')
```


packages/react/src/dashboard/urlFor.ts
```diff
@@ -0,0 +1,460 @@
...
+}
+
+// URL requires an origin; public output always strips this parsing base.
+const relativeUrlBase = 'https://dashboard.invalid'
+
+/** @public */
+export interface EditIntentParameters {
+  id: string
+  type?: string
+  mode?: string
```
cngonzalez: this was the only param i was unsure of the meaning of. i'm not sure how often other teams will be reaching for this or have visibility into it, but might be worth adding a note that it means "edit in structure tool" or "edit in presentation tool"

(if you feel that it's mostly going to be used by people who know this already feel free to ignore this comment)

packages/react/src/dashboard/urlFor.ts
```diff
@@ -0,0 +1,460 @@
...
+  protected edit(update: (url: URL) => void): this {
+    const url = new URL(this.#url)
+    update(url)
+    return this.#create(url)
+  }
+
+  /**
+   * Returns another builder type rooted at the appended path segments.
+   *
+   * Use this when a route moves into a grammar with a different set of available methods.
```
cngonzalez: Is it worth noting when this should be used? I worry about people using it and making something a bit fragile

cngonzalez (dismissed):
exciting foundation! 2 comments on clarification but not blocking

packages/react/src/dashboard/urlFor.ts
gu-stav: Good call - made the type stricter and added a comment 👍🏼

packages/react/src/dashboard/urlFor.ts
gu-stav: Reworded the comment.

joshuaellis:
with the emergence of new singletons should we refrain from adding them specifically and instead using `installation` generally to target something like ML or Persona? Could be a follow up one day, just interested in perspective.

gu-stav:
@joshuaellis I see this as a conceptual wrapper around the routing layer in workbench rather than the data-structures, which is why it does not have a concept of an installation. My thinking is that from user pov I want to build a URL to e.g. a studio document with a perspective without having to know how the URL has to be built.

I am not sure how this is going to evolve - it depends on the popularity on each of these apps and how far we want the support to go. I would only introduce custom grammar for widely used apps.

Making the builder extensible is a good stop-gap solution I think, because if e.g. singleton A is supported by content agent, the team could start to build in the builder there and once the app has gained more popularity and users want to navigate there from their SDK apps or studio, we could move it into core.

packages/react/src/dashboard/urlFor.ts
```diff
@@ -0,0 +1,463 @@
+/** @public */
+export interface Url {
```
ryanbonial: I am concerned that this `Url` may get confused with `URL`, could we name it  `DashboardUrl`. SDK developers might accidentally import our `Url` when writing code for `URL`

packages/react/src/dashboard/urlFor.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    expectTypeOf(
+      urlFor.studios('studio-1').intent('edit', {id: 'document-1'}),
+    ).toEqualTypeOf<StudioIntentUrl>()
+  })
+})
+
+describe('coreApp', () => {
+  it('builds collection and application URLs', () => {
+    expect(urlFor.applications().url()).toBe('/applications')
+    expect(urlFor.applications('app-1').url()).toBe('/application/app-1')
```
ryanbonial: It feels a bit odd that we would have `/applications` (plural) but `/application/app-1` (singular). It seems like it should be `/applications/app-1`

packages/react/src/dashboard/urlFor.ts
gu-stav: Prefixed with "Dashboard" 👍🏼

packages/react/src/dashboard/urlFor.test.ts
gu-stav: We have now aligned on using plural for all routes:

- `/studios` all studios
- `/studios/<id>` single studio
- `/applications` all applications
- `/applicaitons/<id>` single application

Updated the URL builder.

cngonzalez (approved):

PR #1126, opened 2026-08-28
feat(react): port message bus

gu-stav:
### Description

The Dashboard message bus currently lives in `@sanity/workbench`. So independently deployed applications depend on Workbench for the protocol they use to communicate with it.

This ports the message bus into `@sanity/sdk-react`: Workbench remote installs one shared in-memory registry of named topics and applications explicitly connect to it. Registered symbols allow independently bundled SDK copies to find the same registry without importing Workbench.

#### Installing and connecting

Workbench remote owns installation through the internal entrypoint:

```ts
import {installMessageBus} from '@sanity/sdk-react/dashboard/_internal'

const dashboardMessageBus = installMessageBus()
```

Applications connect through the public Dashboard entrypoint:

```ts
import {connectMessageBus} from '@sanity/sdk-react/dashboard'

const messageBus = connectMessageBus()
if (!messageBus) throw new Error('Not connected to Dashboard')
```

`connectMessageBus` returns `undefined` when the host has not installed a bus, the application has no id, or the installed bus is incompatible. It does not install anything or connect at import time.

This removes the side-effect at import-time we previously had when importing `os`.

The export `resetMessageBus` is useful for test isolation. Reset aborts pending work and closes subscriptions while preserving the installed registry, so cached clients remain usable.

#### Topics and operations

All methods operate on a topic name. Each topic contract defines its kind, ownership, payload or state value, and optional reply.

Applications use 3 operations:

- `emit` publishes a state value or sends an event
- `query` returns the current state value, or waits for its first value
- `subscribe` observes state updates or handles events

State topics retain their current value and replay it to new subscribers. An unseeded state waits for its owner to publish; `null` is a published value rather than a missing value.

```ts
// Workbench remote publishes state
dashboardMessageBus.emit('applications.foreground', 'studio')

// Applications read the current value and observe later changes
const foregroundId = await messageBus.query('applications.foreground')

const controller = new AbortController()
messageBus.subscribe(
  'applications.foreground',
  (nextForegroundId) => {
    // React to the new foreground application
  },
  {signal: controller.signal},
)

// Unsubscribe
controller.abort()
```

Event topics deliver occurrences without replay. Responders receive the topic, payload, metadata, reply function, and an abort signal. The metadata travels with the event and identifies the producing application and emission time.

```ts
// Workbench remote responds to a request
dashboardMessageBus.subscribe('auth.token.refresh', (message) => {
  message.reply('new-token')
})

// Applications can await a reply
const token = await messageBus.emit('auth.token.refresh')

// Or send an event without waiting
messageBus.emit('panels.mode.set', {name: 'favorites', mode: 'full'})
```

The event is delivered immediately in both cases. Awaiting the lazy result waits for the first reply and activates `NO_RESPONDER`, `TIMEOUT`, `ABORTED`, and `HANDLER_THREW` failures. Fire-and-forget delivery leaves reply failures unobserved.

#### Compatibility

An application and Workbench remote may bundle different SDK versions, so compatibility has 2 levels:

- The message bus protocol version covers the semantics of `emit`, `query`, `subscribe`, replies, timeouts, and cancellation; changing it is a hard break and clients speaking different protocols do not connect
- Topics evolve independently through adjacent bidirectional migrations; a topic starts at version 1 and its effective version comes from its migration chain

The newer SDK copy supplies the migration chain. State values, event payloads, and event replies are converted between the installed and connected versions:

```ts
{
  from: 1,
  to: 2,
  up: ({name}) => ({fullName: name}),
  down: ({fullName}) => ({name: fullName}),
  reply: {
    up: ({greeting}) => ({salutation: greeting, language: 'en'}),
    down: ({salutation}) => ({greeting: salutation}),
  },
}
```

Topic kind and ownership stay stable across versions. A mismatch rejects the connection instead of leaving the application attached to a contract it cannot interpret.

Ownership controls who publishes state and who responds to events:

- `same_app` means the application that installed the bus publishes the state or responds to the event; every connected application can still send the event
- `any_app` means any connected application can publish the state or respond to the event

#### Available topics

`applications.list` is a partial transport contract. `DashboardApplication` contains `ApplicationBase`, optional `config.mfManifest`, and optional `activeDeployment.interfaces`; interfaces retain the `app`, `panel`, `asset_source`, `worker`, and `tile` variants needed by Dashboard consumers.

State topic values are returned by `query` and replayed by `subscribe`:

- `applications.foreground` - `Application['id'] | null`; ownership `same_app`
- `applications.list` - `TopicResult<DashboardApplication[]> | null`; ownership `same_app`
- `auth.token` - `string | null`; ownership `same_app`
- `applications.config` - `Array<{readonly appId?: Application['id']; readonly appType: 'media-library'; readonly entry: string; readonly moduleId: string; readonly version: string}> | null`; ownership `same_app`
- `navigation.location` - `NavigationLocation | null`; ownership `same_app`
- `organizations.current` - `Pick<OrganizationBase, 'id' | 'name' | 'slug'> | null`; ownership `same_app`
- `panels.mode` - `TopicResult<{appId: string; name: string; mode: 'aside'; size?: number} | {appId: string; name: string; mode: 'full'} | null> | null`; ownership `any_app`
- `preferences.color-scheme` - `'light' | 'dark'`; ownership `any_app`
- `preferences.dock-locked` - `boolean`; ownership `any_app`
- `users.current` - `CurrentUser | null`; ownership `same_app`

Event topics declare the payload sent by `emit` and the reply returned when awaited:

- `auth.token.refresh` - payload `void`; reply `string`; ownership `same_app`
- `navigation.location.update` - payload `{url: string; history?: 'push' | 'replace'}`; reply `{ok: true} | {ok: false; reason: 'not-navigable' | 'interrupted' | 'failed'}`; ownership `same_app`
- `panels.mode.set` - payload `{name: string; mode: 'aside' | 'full'} | {name: string; size: number} | null`; no reply; ownership `same_app`

### What to review

- Whether explicit connection and internal installation form the right package interface
- Whether state, event, ownership, and compatibility semantics preserve the Workbench contract for independently deployed applications
- Whether the initial Dashboard topic contracts are precise enough to maintain over time


packages/react/package.json
```diff
@@ -112,6 +115,7 @@
   "browserslist": "extends @sanity/browserslist-config",
   "packageManager": "pnpm@10.33.4",
   "inlinedDependencies": {
-    "@sanity/comlink": "4.0.3"
+    "@sanity/comlink": "4.0.3",
+    "xstate": "5.32.5"
```
joshuaellis: do we still need this? I don't see where its used now.

joshuaellis (dismissed):
this looks like a 1:1 copy over of the actual bus code

packages/react/package.json
gu-stav: It is still needed, because the package exports types from `@sanity/comlink` which are derived from xstate and must be part of the generated declaration bundle.

joshuaellis: Oh wow TIL. But we didn't have it before though? Is my wondering?

gu-stav: Before it has been a regular dependency, but we can't set it as that, because nothing imports from it. Before it was a peer dependency of `@sanity/workbench`

joshuaellis (dismissed):

joshuaellis (approved):

PR #1127, opened 2026-08-28
feat(react): add `useTopic`

gu-stav:
### Description

React consumers need a typed way to read dashboard message bus topics without wiring subscriptions, Suspense, and topic failures themselves: `useTopic` exposes one React data contract for state values and event payloads.

The hook suspends by default. Pending reads suspend, and a failed `TopicResult` state read throws `TopicError` to the nearest error boundary. Successful state results expose their `value` as `data`.

```tsx
import {ErrorBoundary} from 'react-error-boundary'
import {useTopic} from '@sanity/sdk-react/dashboard'

function Applications() {
  const {data: applications} = useTopic('applications.list')
  return <span>{applications?.length ?? 0} applications</span>
}

  return (
    <ErrorBoundary fallback={<span>Could not load applications</span>}>
      <Suspense fallback={<span>Loading applications</span>}>
        <Applications />
      </Suspense>
    </ErrorBoundary>
  )
}
```

Passing `{suspend: false}` renders immediately and returns a discriminated result:

- `{data: undefined, isPending: true}` before the first value or payload
- `{data: T, isPending: false}` once ready, including when `T` is `null`
- `{error: TopicError, isPending: false}` when a `TopicResult` state read fails; `data` is absent

The failed result remains subscribed, so a later successful value recovers without remounting the consumer.

```tsx
function ApplicationStatus() {
  const applications = useTopic('applications.list', {suspend: false})

  if (applications.isPending) return <span>Loading applications</span>
  if (applications.error) return <span>Could not load applications</span>
  return <span>{applications.data?.length ?? 0} applications</span>
}
```

State topics return their current value and follow later updates. Their initial read uses the message bus query deadline.

Event topics return the latest payload received while consumers are mounted. Their initial read waits for the next payload, and the payload is forgotten after the last consumer unmounts. Result-shaped event payloads remain unchanged.

[SDK-2340](https://linear.app/sanity/issue/SDK-2340)

### What to review

- The Suspense and `suspend: false` result contracts
- State `TopicResult` success, failure, and recovery behavior
- State query deadline behavior
- Event payload lifetime and result-shaped payload handling


packages/react/src/hooks/dashboard/useTopic.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+      return useTopic('auth.token').data
+    }
+
+    await act(async () => {
+      render(
+        <Suspense fallback="Loading">
+          <Token />
+        </Suspense>,
+      )
+    })
```
joshuaellis, line 67: is that actually the way to do it with testing-library/react? 🤮

packages/react/src/hooks/dashboard/useTopic.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+    expect(screen.getByText('Loading')).toBeInTheDocument()
+    expect(query).toHaveBeenCalledWith('auth.token')
+
+    await act(async () => {
+      mocks.client?.emit('auth.token', 'token')
+      await firstValue
```
joshuaellis: should we await this? shouldn't the findByTextbe enough?

joshuaellis:
Superseded by #1212, which rebuilds `useTopic` on the per-instance message bus store from #1200 and the SDK's `createStateSourceHook` pattern.

PR #1128, opened 2026-08-28
feat(react): add `useApplicationForegroundId`

gu-stav:
### Description

Dashboard consumers need a typed way to identify the foreground application without subscribing to an internal topic directly. `useApplicationForegroundId()` returns `Application['id'] | null`, where `null` means Dashboard has no foreground application. It suspends until Dashboard publishes.

```tsx
import {useApplicationForegroundId} from '@sanity/sdk-react/dashboard'

function ForegroundApplication() {
  const foregroundId = useApplicationForegroundId()
  return <span>{foregroundId ?? 'No foreground application'}</span>
}
```


joshuaellis (dismissed):

packages/react/src/hooks/dashboard/useApplicationForegroundId.ts
```diff
@@ -0,0 +1,15 @@
+import {type Application} from '@sanity/sdk'
+
+import {useTopic, type UseTopicOptions, type UseTopicResult} from './useTopic'
+
+/**
+ * Returns the id of the application in the foreground.
+ *
+ * Pass `{suspend: false}` to receive a pending result instead of suspending.
+ * @public
```
cngonzalez, line 18: @\public or @\internal?

ryanbonial: my 2 cents, seems like `@public` is OK here since it gets exported under `@sanity/sdk-react/dashboard`.

ryanbonial (dismissed):

packages/react/src/hooks/dashboard/useApplicationForegroundId.ts
joshuaellis: agree here.

joshuaellis (dismissed):

packages/react/src/hooks/dashboard/useApplicationForegroundId.ts
```diff
@@ -0,0 +1,22 @@
+import {type Application} from '@sanity/sdk'
+
+import {useTopic} from './useTopic'
+
+/**
+ * Returns the id of the application in the foreground, or `null` when Dashboard has none.
+ *
+ * Suspends until Dashboard publishes the foreground application.
```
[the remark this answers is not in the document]
joshuaellis: made the hook suspend.

joshuaellis (approved):

PR #1143, opened 2026-08-31
docs(core): clarify comments in the document store

ryanbonial:
### Description

Comments in `packages/core/src/document` had drifted: padded phrasing, stale references, and a few outright errors. This directory is read on nearly every task that touches the document store, so that cost is paid over and over, by people and by agents.

This rewrites them for clarity. No code changed. The diff is 349 comment lines and 0 non-comment lines.

Two things deliberately not done. Comments were not deleted, only reworded in place. And comments that contradict the code were left as they are rather than quietly corrected, since the code may be the thing that is wrong. Those are listed below.

### What to review

Three comments contradict the code. Each needs a decision, not a rewrite:

- `processMutations.ts` and `processMutations.test.ts` say documents "go into `applyMutations`". No such function exists. The only `applyMutations` is a local variable in `edit.ts` and `shared.ts`, so this looks like a rename that left the comments behind.
- `listen.test.ts` documents `@param rev - The revision (can be undefined)` against a `rev: string` signature, and all ten call sites pass a string.
- `resourceRules.ts` has an em dash in a user-facing `console.warn` string. Left alone because it is a string literal, but flagging it as user-facing copy.

Corrections where I did change the comment, worth a second look:

- `unpublish.ts` said "Standard draft/published or version logic", but the guard eight lines above throws for release perspectives, so the version case is unreachable there. It appears copy-pasted from `discard.ts`, where the same comment is correct.
- `permissions.ts` said a record's values "will be `undefined`" where the cast directly below asserts the opposite.
- `sharedListener.test.ts` named `getSubscribableClient`; the mock is `getClientState`.
- Typos: `_updateAt`, "Medial Library", "frest", "a set of document".

Every `@param`, `@returns`, and release tag is untouched. TypeDoc publishes those, so they are API reference rather than internal prose.

### Testing

No behavior change, so no new tests. The existing suite was run as a guard:

- `pnpm test` scoped to `src/document`: 446 passing
- `pnpm ts:check`, `pnpm lint`, `pnpm fallow audit`, and `oxfmt --check`: clean

### Fun gif
writer


packages/core/src/document/applyDocumentActions.test.ts
```diff
@@ -82,26 +82,21 @@ describe('applyDocumentActions', () => {
...
       base: {doc1: {...exampleDoc, _id: 'doc1', foo: 'old', _rev: 'rev-old'}},
       working: {doc1: {...exampleDoc, _id: 'doc1', foo: 'bar', _rev: 'rev-new'}},
       previous: {doc1: {...exampleDoc, _id: 'doc1', foo: 'old', _rev: 'rev-old'}},
       previousRevs: {doc1: 'rev-old'},
       timestamp: new Date().toISOString(),
     }
-    // Update the state so that its "applied" array contains our applied transaction.
     state.set('simulateApplied', {applied: [appliedTx]})
 
-    // Await the resolution of applyDocumentActions. This should pick up the applied transaction.
```
cngonzalez: I think sometimes it's useful to leave comments that assert the desired / expected behavior?

packages/core/src/document/documentStore.ts
```diff
@@ -105,8 +105,8 @@ export interface DocumentStoreState {
 export interface DocumentState {
   id: string
   /**
-   * the "remote" local copy that matches the server. represents the last known
-   * server state. this gets updated every time we confirm remote patches
+   * the local copy that matches the server: the last known server state,
```
cngonzalez: these both read equally clear to me? if anything, the new version is more LLMy

packages/core/src/document/listen.test.ts
```diff
@@ -153,7 +153,7 @@ describe('sortListenerEvents operator', () => {
   })
 
   it('should throw MaxBufferExceededError when the buffer exceeds the max size', async () => {
-    // Set a very low maxBufferSize (e.g. 2)
+    // maxBufferSize is 2 below, so the three unchainable mutations overflow it
```
cngonzalez: 2 below what?

PR #1144, opened 2026-09-01
feat(react): add `useApplications` and `useApplication`

gu-stav:
### Description

`applications.list` exposes a Dashboard transport payload: a partial Brett application shape with deployment and config fields, while consumers need stable SDK-owned applications with loadable interfaces.

This adds `useApplications` and `useApplication` to `@sanity/sdk-react/dashboard`. The hooks strip the transport-only fields and return `ApplicationBase` plus `views` and `webWorkers`.

An `app` interface becomes a view with `surface: 'window'`. Other non-worker interfaces keep their type as `surface`, while workers become `webWorkers`. Each mapped interface receives a `RemoteModuleRef` with:

- `entry` pointing at the application origin
- `moduleId` prefixed with the application ID
- `version` copied from the interface

When `externalUrl` is present, its origin hosts the module. Otherwise singleton applications resolve to `https://<slug>-apps-<org-id>.sanity.run` and other applications to `https://<slug>.sanity.studio`, with the `.sanity.work` equivalents in staging. An application without `config.mfManifest` is non-federated and returns empty `views` and `webWorkers`.

Both hooks suspend until `applications.list` publishes. A `null` topic value becomes an empty list, a failed result throws `TopicError` to the nearest error boundary, and `useApplication` returns `null` when the requested ID is missing.

```tsx
import {useApplications} from '@sanity/sdk-react/dashboard'

function Applications() {
  const applications = useApplications()

  return applications.map((application) => (
    <div key={application.id}>{application.title}</div>
  ))
}

  return (
    <Suspense fallback={<p>Loading applications...</p>}>
      <Applications />
    </Suspense>
  )
}
```

```tsx
import {useApplication} from '@sanity/sdk-react/dashboard'

function SelectedApplication({id}: {id: string}) {
  const application = useApplication(id)

  return application ? <h1>{application.title}</h1> : <p>Application unavailable</p>
}
```


joshuaellis (dismissed):

joshuaellis:
@claude action point 5

joshuaellis:
Re the review at 979b6357, now 8bf4a98f:

1. The `globalThis` read is deliberate. A remote is built once and runs in whichever host page loaded it, and the host sets the flag at runtime; a bundler `define` would lock the chunk to the environment it was built in. This mirrors workbench's `getSanityEnv` (`core/env.ts`), which documents the same reasoning. `getStagingApiHost` in core is the one with the build-time behaviour, and that's a separate concern. Comment added.

2. Agreed on the throw. An application with interfaces but no `slug` or `externalUrl` now yields empty `views`/`webWorkers`, the same as a non-federated one, with a test. `externalUrl` is typed `string | null` on `ApplicationBase`, so `!== null` stands.

3. No non-React consumer needs this today; it can move to core when one does.

4. Renamed to `DashboardApplication`, `DashboardView`, `DashboardWebWorker`.

5. Added the unresolvable-origin case and `tile`/`asset_source` surface assertions. `useApplication`'s stability falls out of the `useMemo` in `useApplications`.

Minor: comment on the cast added.

joshuaellis (approved):

PR #1145, opened 2026-09-01
feat(react): export DashboardTokenRefreshProvider

skogsmaskin:
## Summary

Makes `DashboardTokenRefreshProvider` a documented public export of `@sanity/sdk-react` (it was `@internal` and only mounted inside `AuthBoundary`).

## Why

Apps that run as federated remotes inside the dashboard (e.g. the Media Library under the workbench shell) can't mount `AuthBoundary`: the shell owns the session, so the app renders its own loading/error UI and must not show the SDK's login flow. But such apps still use SDK hooks (`useQuery` etc.), which read their token from the instance's auth store — and `DashboardTokenRefreshProvider` is exactly the piece that keeps that store in sync with the shell's session (mirrors the `auth.token` topic in, requests a reissue on 401).

With it unexported, the only options were mounting all of `AuthBoundary` or reimplementing the provider with `setAuthToken` — the Media Library currently does the latter, and this PR lets us delete that copy.

Discussed with @joshuaellis in [this Slack thread](https://sanity-io.slack.com/archives/C0BH33BBVGX/p1788182900660489), where the suggestion was to PR it. Context: the provider was deliberately left internal in #1039; this is the concrete downstream use case for exporting it.

## Notes

- `@public` matches its sibling `ComlinkTokenRefreshProvider`, which is already exported.
- No behavior change — docs and export surface only.
- The existing test (`DashboardTokenRefresh.test.tsx`) already mounts the provider without `AuthBoundary`, covering exactly the newly supported usage.

## Test plan

- [x] `pnpm --filter @sanity/sdk-react test` (378 passed)
- [x] `pnpm --filter @sanity/sdk-react ts:check`, `lint` (two pre-existing warnings on main, none new)
- [x] `pnpm --filter "@sanity/sdk-react..." build` (strict + check) — export present in `dist/index.js` / `dist/index.d.ts`
- [x] `pnpm fallow dead-code --baseline fallow-baselines/dead-code.json` — no new findings

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/react/src/_exports/sdk-react.ts
+++ b/packages/react/src/_exports/sdk-react.ts
@@ -6,6 +6,7 @@ export {SanityApp, type SanityAppProps} from '../components/SanityApp'
 export {SDKProvider, type SDKProviderProps} from '../components/SDKProvider'
 export {type DocumentHandle, type DocumentTypeHandle, type ResourceHandle} from '../config/handles'
 export {ComlinkTokenRefreshProvider} from '../context/ComlinkTokenRefresh'
+export {DashboardTokenRefreshProvider} from '../context/DashboardTokenRefresh'
 export {renderSanityApp} from '../context/renderSanityApp'
 export {ResourceProvider, type ResourceProviderProps} from '../context/ResourceProvider'
 export {
--- a/packages/react/src/_exports/sdk-react.ts
+++ b/packages/react/src/_exports/sdk-react.ts
@@ -6,7 +6,6 @@ export {SanityApp, type SanityAppProps} from '../components/SanityApp'
 export {SDKProvider, type SDKProviderProps} from '../components/SDKProvider'
 export {type DocumentHandle, type DocumentTypeHandle, type ResourceHandle} from '../config/handles'
 export {ComlinkTokenRefreshProvider} from '../context/ComlinkTokenRefresh'
-export {DashboardTokenRefreshProvider} from '../context/DashboardTokenRefresh'
 export {renderSanityApp} from '../context/renderSanityApp'
 export {ResourceProvider, type ResourceProviderProps} from '../context/ResourceProvider'
 export {
--- a/packages/react/src/context/DashboardTokenRefresh.tsx
+++ b/packages/react/src/context/DashboardTokenRefresh.tsx
@@ -70,14 +70,15 @@ function DashboardTokenRefresh({children}: PropsWithChildren) {
  *
  * @example
  * ```tsx
- * import {DashboardTokenRefreshProvider, ResourceProvider} from '@sanity/sdk-react'
+ * import {ResourceProvider} from '@sanity/sdk-react'
+ * import {TokenRefreshProvider} from '@sanity/sdk-react/dashboard'
  *
  * function EmbeddedApp() {
  *   return (
  *     <ResourceProvider fallback={<Loading />}>
- *       <DashboardTokenRefreshProvider>
+ *       <TokenRefreshProvider>
  *         <App />
- *       </DashboardTokenRefreshProvider>
+ *       </TokenRefreshProvider>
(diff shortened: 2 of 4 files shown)


packages/react/src/context/DashboardTokenRefresh.tsx
```diff
@@ -48,9 +48,38 @@ function DashboardTokenRefresh({children}: PropsWithChildren) {
 }
 
 /**
- * Provides dashboard OS token refresh. No-op outside the dashboard, where the
- * app uses its normal auth flow.
- * @internal
+ * Keeps the SDK's auth store in sync with the session owned by the Sanity
+ * Dashboard when the app runs as a federated remote inside it. Renders
+ * children unchanged outside the dashboard, where the app uses its normal
+ * auth flow.
```
joshuaellis: this comment doesn't actually say anything, can you rewrite please.

packages/react/src/_exports/sdk-react.ts
```diff
@@ -6,6 +6,7 @@ export {SanityApp, type SanityAppProps} from '../components/SanityApp'
 export {SDKProvider, type SDKProviderProps} from '../components/SDKProvider'
 export {type DocumentHandle, type DocumentTypeHandle, type ResourceHandle} from '../config/handles'
 export {ComlinkTokenRefreshProvider} from '../context/ComlinkTokenRefresh'
+export {DashboardTokenRefreshProvider} from '../context/DashboardTokenRefresh'
```
joshuaellis: @gu-stav should this be exported from the `dashboard` sub export path? I would assume so.

packages/react/src/context/DashboardTokenRefresh.tsx
```diff
@@ -48,9 +48,38 @@ function DashboardTokenRefresh({children}: PropsWithChildren) {
...
+ *
+ * @remarks
+ * `AuthBoundary` mounts this automatically, so most apps never need it
+ * directly. Mount it yourself when your app runs inside the dashboard but
+ * cannot use `AuthBoundary` — for example a federated app that renders its
```
joshuaellis: don't discuss "federation" its an implementation detail

joshuaellis (changes requested):

packages/react/src/context/DashboardTokenRefresh.tsx
skogsmaskin: Rewrote it — the summary now says what the provider actually does (subscribes to the dashboard-issued token, writes each value into the auth store, requests a fresh token on 401, renders children unchanged outside the dashboard).

packages/react/src/context/DashboardTokenRefresh.tsx
skogsmaskin: Dropped — the docs now only talk about running inside the dashboard, no federation.

packages/react/src/_exports/sdk-react.ts
gu-stav: Yes, I think so and without the "dashboard" prefix:

```
TokenRefreshProvider
```

skogsmaskin: Done — exported as `TokenRefreshProvider` from `@sanity/sdk-react/dashboard` (removed from the main entry), docs example updated to match.

skogsmaskin:
Follow-up sweep in f0d7f12: dropped the remaining "federated remote"/"federation" wording from the internal docs too (the `DashboardTokenRefresh` function doc and three spots in `dashboardToken.ts`), so the internal comments now match the public docs. They describe the setup as apps embedded in the dashboard's own window — keeping the detail that matters (same realm as the host, unlike the Core UI iframe).

joshuaellis (approved):

PR #1147, opened 2026-09-01
feat(react): adapt dashboard hooks to the message bus

gu-stav:
### Description

Dashboard hooks only read the Comlink-backed SDK state, so federated applications cannot receive authentication, user, organization, or navigation updates from Dashboard.

This supports both Dashboard runtime paths: embedded applications keep using the existing Comlink-backed SDK state, while federated applications use message bus topics.

Existing consumer calls stay unchanged:

```tsx
const token = useAuthToken()
const user = useCurrentUser()
const organizationId = useOrganizationId()

useNavigate(({path, type}) => {
  router.navigate(path, {replace: type === 'replace'})
})
```

`users.current` now carries `CurrentUser | null`, so `useCurrentUser` keeps its full return contract. SDK state remains the initial value until an unseeded bus topic publishes, preserving the hooks' non-suspending API.

[SDK-2317](https://linear.app/sanity/issue/SDK-2317/adapt-existing-dashboard-hooks-to-the-message-bus)

### What to review

- Runtime selection between Comlink-backed SDK state and message bus topics
- Initial state fallback before a federated topic publishes
- Navigation filtering for the foreground federated application


joshuaellis:
Superseded by SDK-2314 (message bus, useTopic, useEmit, useApplication* are on main). Remaining scope is tracked in SDK-2578 and SDK-2579.

PR #1159, opened 2026-09-02
fix(documents): write only genuinely changed fields in useEditDocument

ryanbonial:
### Description

Full-document edits now compare values deeply instead of by reference, so `useEditDocument` writes only the fields whose contents changed. Path edits skip the write when the next value already equals the current one. The cost is a deep comparison of every top-level field on each full-document write, unmeasured, bounded by document size.

The hook has always documented that it emits "the minimal set of `set` and `unset` operations required for the update", but full-document mode compared top-level keys with `!==`. A caller that re-parses or clones its value hands back a fresh reference for every object and array, so the filter emitted a whole-field `set` for fields that were byte-identical. Path mode read the current value and then wrote unconditionally.

A whole-field `set` on a Portable Text field overwrites the operational patches beneath it, and the editor typing in that field watches their text revert. Offline that costs everything queued for the field, not just the last flush.

Both branches already read the current value, so the comparison adds no reads.

An edit that changes nothing produces a transaction with zero actions, and that exposed a latent bug in core. `batchAppliedTransactions` in `reducers.ts` skips empty transactions, so their ids never reach `batchedTransactionIds`, the `accepted` event never matches, and the transaction sits in `applied` with a `submitted()` promise that never resolves. `applyDocumentActions` now returns a no-op result before the transaction reaches the queue.

Behavior change: a caller that re-set an identical value to bump `_updatedAt` no longer gets a write.

Ticket: https://linear.app/sanity/issue/SDK-2475

### What to review

`packages/react/src/hooks/document/useEditDocument.ts` holds both comparisons. Neither no-op path returns early; both call `apply` with an empty action list, so the caller still gets an `ActionsResult`.

`packages/core/src/document/applyDocumentActions.ts` holds the guard, and the TSDoc on `emptyActionsResult` records why an empty transaction cannot be queued.

Both `@sanity/sdk-react` and `@sanity/sdk` change, so both take a patch.

Open question for reviewers: whether the deep comparison should be skipped above some document size. I have no measurement either way.

### Testing

Five new unit tests: a re-parsed document writes only genuinely changed keys, a no-op full-document write emits nothing, a no-op path write emits nothing, an empty transaction resolves without queueing, and `useApplyActions` forwards an empty action list with its resource. I reverted the fix and confirmed all five fail without it.

Checked by hand in the kitchensink JSON editor against a live document. Editing `role` still writes and syncs. Logging each key's comparison showed `role` as the only reference-unequal key, with the seven-block Portable Text array and every other field reference-equal, so `json-edit-react` preserves references for branches you do not touch and its per-value edits never hit the clone path.

### Fun gif
nothing


binoy14 (commented):
Looks okay from code perspective but don't have too much context, would be good if @cngonzalez could take a peek as well

cngonzalez (approved):
nice!

PR #1162, opened 2026-09-03
feat(core): add OAuth authorization-code + PKCE auth mode

joshuaellis:
### Description

Adds `oauth` as a fourth auth mode in `packages/core`, alongside studio, dashboard, and standalone. Configured via `auth.oauth` (`clientId`, `redirectUri`, `organizationId`), it uses the OAuth 2.0 authorization-code + PKCE flow and scopes tokens to an organisation via RFC 8707 resource indicators.

Core handles PKCE generation, the authorize redirect, code exchange, token refresh, revocation, and cross-tab token persistence, exposed as bound actions (`startOAuthAuthorization`, `handleOAuthCallback`, `refreshOAuthTokens`, `revokeOAuthTokens`) and a `getOAuthTokensState` accessor. `handleOAuthCallback` is single-flight so a repeated or StrictMode-double-invoked call does not attempt a second exchange of a single-use code.

### Fun Gif

gif-by-lakeshore-records

### DIFF
--- a/packages/core/src/auth/oauth/oauthActions.ts
+++ b/packages/core/src/auth/oauth/oauthActions.ts
@@ -1,0 +1,371 @@
+import {type SanityClient} from '@sanity/client'
+
+import {bindActionGlobally} from '../../store/createActionBinder'
+import {createStateSourceAction} from '../../store/createStateSourceAction'
+import {type StoreContext} from '../../store/defineStore'
+import {DEFAULT_BASE, REQUEST_TAG_PREFIX} from '../authConstants'
+import {getAuthLogger} from '../authLogger'
+import {AuthStateType} from '../authStateType'
+import {authStore, type AuthStoreState} from '../authStore'
+import {createLoggedInAuthState, getDefaultLocation} from '../utils'
+import {generateCodeChallenge, generateCodeVerifier, generateState} from './pkce'
+import {type OAuthTokens} from './types'
+
+/** sessionStorage key for the PKCE `code_verifier`. */
+export const OAUTH_VERIFIER_KEY = '__sanity_oauth_verifier'
+
+/** sessionStorage key for the CSRF `state` value. */
+export const OAUTH_STATE_KEY = '__sanity_oauth_state'
+
+interface TokenEndpointResponse {
+  access_token: string
+  token_type: string
+  expires_in: number
+  refresh_token?: string
+}
+
+/** Builds the RFC 8707 resource indicator for an organisation. */
+function getResourceIndicator(organizationId: string): string {
+  return `urn:io.sanity:organization:${organizationId}`
+}
+
+/**
+ * Serialises tokens for storage, converting `expiresAt` to an ISO string.
+ *
+ * @internal
+ */
+export function serializeTokens(tokens: OAuthTokens): string {
+  return JSON.stringify({
+    accessToken: tokens.accessToken,
+    tokenType: tokens.tokenType,
+    expiresIn: tokens.expiresIn,
+    expiresAt: tokens.expiresAt.toISOString(),
+    ...(tokens.refreshToken !== undefined && {refreshToken: tokens.refreshToken}),
+  })
+}
+
+type AuthOptions = AuthStoreState['options']
+type ConfiguredOAuthOptions = AuthOptions & {oauth: NonNullable<AuthOptions['oauth']>}
+
+/**
+ * Reads the store options, throwing when the instance was not configured for
+ * OAuth.
+ */
+function getOAuthOptions(state: AuthStoreState): ConfiguredOAuthOptions {
+  const {options} = state
+  if (!options.oauth) {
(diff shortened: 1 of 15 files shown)


joshuaellis:
i think the refresh logic will go away and the client will [handle it](https://github.com/sanity-io/client/pull/1323) but to be safe and because code is cheap, its added here.

Although if I do delete it, it'll be a BC to the SDK right? so maybe I should just mark it as `@experimental`? 😅

packages/core/src/auth/oauth/oauthActions.ts
```diff
@@ -0,0 +1,371 @@
...
+  const logger = getAuthLogger(instance)
+  const options = getOAuthOptions(state.get())
+
+  const codeVerifier = generateCodeVerifier()
+  const oauthState = generateState()
+  const codeChallenge = await generateCodeChallenge(codeVerifier)
+
+  const session = typeof sessionStorage !== 'undefined' ? sessionStorage : undefined
+  session?.setItem(OAUTH_VERIFIER_KEY, codeVerifier)
+  session?.setItem(OAUTH_STATE_KEY, oauthState)
```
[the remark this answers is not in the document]
joshuaellis: 👀  @claude what would be the recommended fix here?

packages/core/src/auth/oauth/oauthAuth.ts
```diff
@@ -0,0 +1,168 @@
...
+import {DEFAULT_BASE} from '../authConstants'
+import {AuthStateType} from '../authStateType'
+import {type AuthStoreState} from '../authStore'
+import {type AuthStrategyOptions, type AuthStrategyResult} from '../authStrategy'
+import {subscribeToStateAndFetchCurrentUser} from '../subscribeToStateAndFetchCurrentUser'
+import {createLoggedInAuthState, getDefaultStorage, getStorageEvents} from '../utils'
+import {type OAuthTokens} from './types'
+
+/** localStorage (or configured `storageArea`) key for persisted OAuth tokens. */
+export const OAUTH_TOKENS_KEY = '__sanity_oauth_tokens'
```
ryanbonial: [blocking] What if there are multiple SDK apps that use different OAuth? The tokens could be sent to the wrong OAuth server

packages/core/src/auth/oauth/oauthActions.ts
```diff
@@ -0,0 +1,371 @@
...
+      state.set('oauthExchangeError', {
+        authState: {type: AuthStateType.ERROR, error: exchangeError},
+      })
+      return cleanedUrl
+    }
+  },
+)
+
+// Single-flight refresh shared across concurrent callers. Safe as a module
+// singleton because `authStore` is a global store (one shared state).
```
ryanbonial, line 253: Heads up, this is really one shared state per tab, so it could be possible (although unlikely) for two tabs with the same app loaded to race for a token refresh.

packages/core/src/auth/oauth/oauthActions.ts
```diff
@@ -0,0 +1,371 @@
...
+})
+
+async function doRefreshOAuthTokens({
+  state,
+  instance,
+}: StoreContext<AuthStoreState>): Promise<OAuthTokens | null> {
+  const logger = getAuthLogger(instance)
+  const options = getOAuthOptions(state.get())
+
+  const current = state.get().oauthTokens
```
ryanbonial, line 282: [blocking] I don't see anywhere in the `logout()` that these state `oauthTokens` are being removed from the store

packages/core/src/auth/oauth/oauthActions.ts
```diff
@@ -0,0 +1,371 @@
...
+    options.storageArea?.setItem(options.storageKey, serializeTokens(tokens))
+    logger.info('OAuth tokens refreshed')
+    state.set('oauthRefreshed', {
+      authState: createLoggedInAuthState(tokens.accessToken, null),
+      oauthTokens: tokens,
+    })
+    return tokens
+  } catch (error) {
+    logger.error('OAuth token refresh failed — logging out', {error})
+    options.storageArea?.removeItem(options.storageKey)
```
ryanbonial, line 327: A full logout seem aggressive here for something that may just be a network issue, but not a blocker.

ryanbonial (changes requested):

packages/core/src/auth/oauth/oauthActions.ts
joshuaellis: 🤔 good point. I wonder how other oauth clients handle this.

packages/core/src/auth/oauth/oauthActions.ts
joshuaellis: updated.

packages/core/src/auth/oauth/oauthActions.ts
joshuaellis: updated.

ryanbonial (approved):

PR #1179, opened 2026-09-03
feat(typegen): drop the experimental groq fork behind a compatibility shim

### LINKED ISSUE #1138: Prerelease publishes lack npm provenance; stable @sanity/sdk pins an unattested groq prerelease
## Summary

`@sanity/sdk` (every release from 2.14.0 through 3.0.0) hard-pins `groq@3.88.1-typegen-experimental.0`, a prerelease that was published **without npm provenance**, while every stable `groq` release around it (3.29.0, 6.11.0, …) is attested. The `3.0.0-rc.1` / `3.0.0-rc.2` prereleases of `@sanity/sdk` itself were also published without provenance, while every stable `@sanity/sdk` version is attested.

Installers that enforce a no-downgrade provenance policy refuse to install anything that depends on `@sanity/sdk` — including `sanity@6` — because a package that once shipped provenance is now resolving to a version that ships none.

The ask here is about publishing hygiene: the provenance signal that every stable release already carries is missing on the prerelease path, and stable releases pin directly into that path.

## What happens

Installing `sanity@6` via [mise](https://mise.jdx.dev) (which uses [aube](https://aube.jdx.dev/security#trust-policy) with the default `trustPolicy: no-downgrade`) fails with:

```
trust downgrade for groq@3.88.1-typegen-experimental.0 (trustPolicy=no-downgrade):
earlier published version 3.29.0 had provenance attestation but this version has no trust evidence
```

After excluding that package, the same error appears for `@sanity/sdk@3.0.0-rc.1` and `@sanity/sdk@3.0.0-rc.2`.

Users of npm/pnpm/yarn/bun don't see this today because those clients don't verify provenance continuity by default — but the provenance gap is the same for everyone, and npm is moving toward surfacing it.

## Evidence (from the public registry, 2026-08-28)

Dependency chain from the current stable `sanity`:

```
sanity@6.11.0            provenance: yes   depends on @sanity/sdk-react@^2.20.1
└ @sanity/sdk-react@2.20.2  provenance: yes   depends on @sanity/sdk@2.20.2
  └ @sanity/sdk@2.20.2      provenance: yes   depends on groq@3.88.1-typegen-experimental.0  (exact pin)
    └ groq@3.88.1-typegen-experimental.0   provenance: NONE
```

| Package | Version | Provenance | Publisher | Published |
|---|---|---|---|---|
| `groq` | 3.29.0 | yes (slsa v0.2) | sanity-io CI | 2024-02-13 |
| `groq` | **3.88.1-typegen-experimental.0** | **none** | `ricokahler` (manual) | 2025-05-07 |
| `groq` | 6.11.0 (latest) | yes (slsa v1) | sanity-io CI | — |
| `@sanity/sdk` | 2.14.0 … 2.20.2 | yes (slsa v1) | `sanity-svc.npm` | — |
| `@sanity/sdk` | **3.0.0-rc.1** | **none** | `sanity-svc.npm` | 2026-07-29 |
| `@sanity/sdk` | **3.0.0-rc.2** | **none** | `sanity-svc.npm` | 2026-08-24 |
| `@sanity/sdk` | 3.0.0 | yes (slsa v1) | `sanity-svc.npm` | 2026-08-27 |

Every `@sanity/sdk` version from 2.14.0 to 3.0.0 (and `@sanity/sdk-react` 2.20.x / 3.0.0) declares `"groq": "3.88.1-typegen-experimental.0"` — an exact pin to a 15-month-old unattested experimental build, reachable from the stable `sanity` package.

Reproduce:

```sh
npm view @sanity/sdk@3.0.0 dependencies.groq dist.attestations
npm view @sanity/sdk@3.0.0-rc.2 dist.attestations       # empty
npm view groq@3.88.1-typegen-experimental.0 dist.attestations   # empty
npm view groq@3.29.0 dist.attestations                   # present
```

## Why this matters

- **Provenance ratchets.** Once a package has shipped attested versions, a later version without attestation is indistinguishable from a compromised publish (leaked token, hijacked account). That is exactly the signal `no-downgrade` policies are designed to catch, and why they can't tell a "legitimate experimental build" from an attack.
- **The pin propagates the gap.** Even though `@sanity/sdk@3.0.0` itself is attested, its exact pin means every downstream install transitively depends on an unattested artifact. The stable SDK effectively inherits the trust level of its weakest pinned dependency.
- **The cost lands on consumers.** Today the workaround is per-version manual verification plus a trust-policy exclusion — every consumer with provenance checking has to re-verify Sanity's publishes by hand, on every fresh resolution, until the pin changes.

## Suggested fixes

1. **Publish prereleases with provenance.** Stable releases are already attested, so the pipeline exists; `npm publish --provenance` works identically for `rc`/experimental dist-tags. Routing prerelease publishes through the same workflow closes the gap for future RCs.
2. **Repin `groq` in `@sanity/sdk` / `@sanity/sdk-react` to an attested version** — either an attested stable `groq` release, or a re-publish of the `typegen-experimental` build through the provenance workflow. Stable SDK releases shouldn't pin exact unattested prereleases at all.

ryanbonial:
### Description

The SDK declares `groq@3.88.1-typegen-experimental.0` as a runtime dependency of both published packages. pnpm 11 with `trustPolicy: no-downgrade` refuses to install it. This breaks installs for SDK apps with that setup, including apps that never asked for TypeGen.

This moves the catalog to released `groq` (`^6.12.0`) and removes React's unnecessary runtime dependency on it. The hooks use SDK-owned aliases for the legacy types. A compatibility shim re-declares the fork's type surface on released `groq`, so existing experimental `sanity.types.ts` files can keep compiling and inferring after the dependency and import changes below.

This is the dependency cleanup ahead of the new Typegen work. `ResolveDocument`, `ResolveQueryResult`, and `ResolveProjectionResult` preserve the legacy lookups. Current Typegen output continues to use explicit hook generics.

Existing experimental users need to replace their direct `groq` dependency with released `groq` and move `defineProjection` imports to `@sanity/sdk` or `@sanity/sdk-react`. Their generated file and the SDK must resolve the same released `groq` declarations. Keeping the fork alongside it can cause declaration conflicts or silently lose inference.

Keep the existing generated file during that migration. The experimental scanner does not recognize `defineProjection` imported from the SDK, and current Typegen does not generate projection types. To resume generation, migrate to current Typegen and pass generated types explicitly to hooks. The guide covers both paths.

Ticket: SDK-2271

### What to review

- Start with [packages/react/guides/Typescript.md](https://github.com/sanity-io/sdk/blob/SDK-2271/typegen-compatibility-shim/packages/react/guides/Typescript.md). It separates current Typegen setup from migration with a saved experimental file, including the dependency and regeneration limitations.
- `groqCompat.ts` preserves the fork's semantics, including `SafeAccess`'s union fallback. `resolve.ts` exposes those lookups through the SDK. The hook changes use those aliases without changing runtime behavior.
- `defineProjection` is an identity function exported by the SDK because released `groq` has no runtime implementation. Existing projection strings must still match their saved generated entries.
- The kitchensink uses explicit projection types after the import change. Its generator script removes helper imports that the experimental generator emits but no longer uses, which otherwise fail the app's typecheck. This is internal tooling, not a customer migration script.
- `scripts/pack-smoke-test.mts` now compiles one legacy consumer against the packed core and React packages with `tsc --noEmit`. The fixture checks document, query, and projection inference through `@sanity/sdk-react`. Fixture files are stored as text so their module augmentations only enter the temporary consumer's TypeScript program.
- Renovate now updates released `groq`. The kitchensink's experimental `@sanity/codegen` remains pinned until its migration.

### Testing

- Local `ts:check`, `lint`, and `test` passed: 1,645 tests passing, 3 skipped. Lint reports 7 existing warnings. Core and React package builds and the Fallow audit passed.
- `defineProjection.test.ts` covers identity return, whitespace preservation, and literal-type preservation. The packed-consumer smoke test passed.
- Kitchensink generation and its typecheck passed with the updated imports and explicit projection types.
- Installed the packed React package and core override in a fresh consumer using pnpm 11.25.0 with `trustPolicy: no-downgrade` and `--ignore-scripts`. It resolved released `groq@6.12.0` without a trust exception.

### Fun gif

shimmy


pnpm-workspace.yaml
```diff
@@ -48,7 +48,7 @@ catalog:
   '@vitejs/plugin-react': '^6.0.3'
   '@vitest/coverage-v8': '^4.1.10'
   eslint: '^10.9.1'
-  groq: '3.88.1-typegen-experimental.0'
+  groq: '~6.12.0'
```
ryanbonial: there it is

apps/kitchensink-react/scripts/typegen.mjs
```diff
@@ -41,6 +41,27 @@ const GENERATED_FILE_WARNING = `/**
  * ---------------------------------------------------------------------------------
  */\n\n`
 
+/**
+ * Removes helper types the generator imports from `groq` but never uses, which trips
+ * `noUnusedLocals` in the app's typecheck. It emits the full helper list unconditionally,
+ * so a run with no projections leaves `ProjectionBase` dangling.
+ */
+function dropUnusedGroqImports(code) {
```
cngonzalez, line 49: is this here because we're considering making it available to users who may need to make the same step? Or is it just something we did to make things work?

apps/kitchensink-react/src/DocumentCollection/DocumentProjectionRoute.tsx
```diff
@@ -1,5 +1,6 @@
+import {defineProjection} from '@sanity/sdk'
 import {DocumentHandle, useDocumentProjection, usePaginatedDocuments} from '@sanity/sdk-react'
-import groq, {defineProjection} from 'groq'
```
cngonzalez, line 2: i know we did this because we wanted to get the experimental groq out of OUR dependencies -- what happens if a user updates their sdk version but keeps the experimental groq in their own deps and continues to import this? does everything still "just work"?

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,235 @@
...
+[Sanity TypeGen](https://www.sanity.io/docs/sanity-typegen) generates TypeScript types
+from your schemas and GROQ queries. The SDK hooks read them, so `useQuery` returns the
+shape your query selects and `useDocument` returns the document type your handle names.
 
-This guide explains how to set up and use Typegen within your SDK project.
+What you get depends on how your app was set up, because the generation side is
+mid-rebuild:
 
-## Benefits
+|                         | Set up with the experimental packages | New app, using `sanity typegen generate` |
```
cngonzalez: This "sanity typegen generate" is assuming we're using the current GA Typegen, correct?

There's some existing docs on migration / mitigation here: https://www.sanity.io/docs/app-sdk/sdk-typegen and https://www.sanity.io/docs/app-sdk/migrating-from-experimental-typegen-in-app-sdk that might be worth a reference and / or syncing up with this.

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+# Using TypeGen with the Sanity SDK (beta)
 
-[Sanity TypeGen](https://www.sanity.io/docs/sanity-typegen) is a tool that generates TypeScript types directly from your Sanity schemas and GROQ queries. When used with the Sanity SDK, it provides strong type safety and autocompletion for your documents, query results, and projections, significantly improving the development experience.
+[Sanity TypeGen](https://www.sanity.io/docs/sanity-typegen) generates TypeScript types
+from your schemas and GROQ queries. The SDK hooks read them, so `useQuery` returns the
+shape your query selects and `useDocument` returns the document type your handle names.
 
-This guide explains how to set up and use Typegen within your SDK project.
+What you get depends on how your app was set up, because the generation side is
+mid-rebuild:
```
cngonzalez:
```suggestion
undergoing changes in anticipation of GA:
```

I personally got a little confused at this phrasing -- like, "does the generation happen in the middle of me rebuilding my app?"

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+| Stable                                  | In beta                                                          |
+| --------------------------------------- | ---------------------------------------------------------------- |
+| The hooks themselves                    | `defineProjection`'s home. It moves from `@sanity/sdk` to `groq` |
+| `defineQuery`, from `groq`              | The `SanitySchemasByResource` interface name                     |
+| Handles and the `create*Handle` helpers | `ResolveDocument` and its siblings                               |
+| What you write in a component           | The bridge file, which gets deleted                              |
 
-First, you need to extract your Sanity schema(s) into a JSON format that Typegen can understand. **Currently, this step relies on the full `sanity` package**, typically used within your Sanity Studio project, as Typegen needs access to the complete schema definition to generate accurate types.
+What you write stays as it is. What changes sits underneath it, so expect an import to
+move and a hand-written file to become unnecessary, not a rewrite of your components.
```
cngonzalez: what hand-written file is being referred to? A hand-written type file?

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+| `useQuery`              | Typed                                 | Typed                                    |
+| `useDocument`           | Typed                                 | Typed, after the bridge file below       |
+| `useDocumentProjection` | Typed                                 | Not yet                                  |
 
-- **Type Safety:** Catch errors at build time instead of runtime.
-- **Autocompletion:** Get intelligent suggestions for document fields and query results in your editor.
-- **Improved Maintainability:** Types make code easier to understand and refactor.
-- **Dataset & Document Scoping:** Generate types that correctly narrow based on the specific dataset or document type context.
+The full setup lives in
+[App SDK and TypeGen](https://www.sanity.io/docs/app-sdk/sdk-typegen). This guide covers
```
cngonzalez: We likely shouldn't link this if further down we say don't install the experimental packages? This article does lead to a migration article but it's not quite the same as starting "fresh"

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+Your app needs the `schema.json` file, not the `sanity` package. If you keep a
+`sanity-typegen.json`, do not set `overloadClientMethods` to `false`; that flag
+suppresses the module augmentation the SDK reads.
 
-The generated file contains types for your schema documents, projections, and query results. The SDK hooks will automatically pick up these types if the Typegen file exists in your project.
+**Your app and the SDK must resolve the same copy of `groq`.** The SDK's compatibility
+declarations attach to the `groq` directory its own types resolve to, so a second copy at
+a different version leaves your generated file importing helpers from a copy that never
+received them, and it fails with `TS2614: Module '"groq"' has no exported member
+'SchemaOrigin'`. The `^6` range above is what keeps them deduped. Pinning `groq` to an
```
cngonzalez: Should we just say people need to install `groq@^6`?

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+  }
 }
 ```
 
-### Handles and Literal Types: `createDocumentHandle` vs `as const`
+Add one entry per dataset. Pass the whole `AllSanitySchemaTypes` union; it includes
+object types such as `slug` alongside your documents, and the SDK filters those out.
+Delete this file once the multi-resource command generates the same declaration.
+
+Three ways to get `never` out of this file, all silent at the declaration:
```
cngonzalez: Why is `never` a concern? Was this something that popped up during the refactor that the agent is compensating for? Would be good to give readers context on why they should care before reading 3 wordy bullet points

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+  dataset: 'production',
   documentId: '123',
   documentType: 'book',
-  dataset: 'production',
-  projectId: 'abc',
 })
 ```
 
-Alternatively, if you prefer defining handles as plain objects, use `as const`:
+A plain object works with `as const`. Prefer the helpers: they are shorter and they fail
```
cngonzalez: Maybe worth keeping the context that you're doing this as part of a document handle (that is, the "plain object" being referred to here IS the document handle).

There's also the consideration that a "true" document handle has extra parameters that MOST utilities don't need (since we resolve to context / defaults for things like projectId / dataset). 

Is that bad for Typegen and we should be enforcing it? Does it play well with resources? (In which case we should revisit and ensure it resolves the right projectId / dataset -- not necessary in THIS pr, just making a note)

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
-  projectId: 'abc',
-} as const // 'as const' ensures documentType is 'book', not string
+import {createDatasetHandle} from '@sanity/sdk'
+import {useQuery} from '@sanity/sdk-react'
+import {defineQuery} from 'groq'
+
+const allBooks = defineQuery('*[_type == "book"]{_id, title}')
+const dataset = createDatasetHandle({projectId: 'abc', dataset: 'production'})
 
-// Now handle.documentType has the literal type 'book'
```
cngonzalez, line 196: Again -- to me it's quite nice to have the comment here explaining why this matters; I may be too basic for the agents, though

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+      {data.map((book) => (
+        <li key={book._id}>{book.title}</li>
+      ))}
+    </ul>
+  )
+}
 ```
 
-**Recommendation:** Use `createDocumentHandle` (or other `create*Handle` helpers) when using Typegen for cleaner code.
+A query that is not a literal, built from a plain string or assembled at runtime, is not
```
cngonzalez: Phrasing is a bit confusing -- this kind of implies that string literals are also looked up? I thought we needed the `groq` or defineQuery helper?

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+looked up. In a new app that yields `never`. In an app carrying a legacy
```
cngonzalez: Again -- a little LLMese-y; I feel the "migrating" and "how Typegen works" sections are a bit confused.

Does the type inference in `useQuery` work differently if you have a types file made from the experimental package vs starting new?

What triggers the "legacy lookup"?

packages/react/guides/Typescript.md
```diff
@@ -1,292 +1,240 @@
...
+  const {data} = useDocumentProjection({...doc, projection: authorSummary})
+  return <span>{data?.name}</span>
 }
 ```
 
-- The generated type (e.g., `AuthorSummaryProjectionResult`) includes a `ProjectionBase` brand, allowing unions of projection results if a projection applies to multiple document types.
-- Typegen intelligently removes types from the projection result if all fields in the projection evaluate to `null` for a given document type.
-- When using Typegen, you **cannot** pass raw projection strings to `useDocumentProjection` and get type inference; you must use `defineProjection`.
+A projection runs against every document type in the schema, so its result narrows by
+document type as well as by dataset. That is why the handle's `documentType` matters here
```
cngonzalez: `documentType` isn't optional in `useProjection` anyway -- not sure if we need to add extra noise here, but I might be misunderstanding the audience.

packages/react/guides/Typescript.md
```diff
@@ -295,71 +243,51 @@ function MixedList() {
...
-- Change your Sanity schemas.
-- Add or modify queries/projections defined with `defineQuery` or `defineProjection`.
-- Consider integrating this into your `dev` script or a file watcher.
-
-### Typegen is additive
-
-Typegen is designed to enhance the SDK experience. If you don't use it, the SDK hooks will still work, but data types will often default to `any` or `unknown`, losing the benefits of TypeScript. Adopting Typegen later should be a non-breaking change that simply adds type safety.
-
-### JavaScript Projects
+The second parameter is required in practice. Omit it and there is no key to match, so a
```
cngonzalez: Required if the same type is in both datasets (relatively rare) or all the time?

packages/react/guides/Typescript.md
```diff
@@ -295,71 +243,51 @@ function MixedList() {
...
 
-While your JavaScript code won't undergo compile-time type checking, modern code editors (like VS Code) that use the TypeScript language service can read this generated file.
+**TypeGen is additive.** Without it, `useQuery` and a literal-`documentType`
+`useDocument` both resolve to `never`, because every lookup misses; pass an explicit
+generic in that case. An untyped handle still gives `useDocument` the base document
+shape. Runtime behavior is identical either way, so it is safe to adopt in an existing
+app and safe to leave out.
 
-This often results in significantly better autocompletion and Intellisense within your JavaScript files when interacting with SDK hooks and data, compared to working without Typegen. Remember, using `defineProjection` and `defineQuery` is still required for Typegen to generate types for those specific artifacts.
+**JavaScript projects benefit too.** Editors read the generated declarations for
```
cngonzalez: We don't really support pure JS projects with SDK -- all init only happens in Typescript. I suppose someone could force it? 🤷

packages/core/scripts/assertAugmentations.mjs
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable no-console -- a build step; its console output is the failure report */
```
cngonzalez: Does this only run in core and not against a customer's build? This feels a bit like overkill to me but I'm open to better understanding why we need to walk through every file for something I thought our bundler did...

cngonzalez (changes requested):
The logic itself looks good -- I got a little confused about seemingly one-off or unusual scripts, but maybe I'm misunderstanding.

I also have some suggestions in the Markdown doc to reduce LLMese and confusion; let me know if I wasn't careful enough in reading it.

apps/kitchensink-react/scripts/typegen.mjs
ryanbonial: This is just to keep our kitchensink generation working, not an example of what we woudl want a customer to do.

apps/kitchensink-react/src/DocumentCollection/DocumentProjectionRoute.tsx
ryanbonial: It would likely be problematic and cause failures. Uses will need to replace the experimental groq dependency with released groq and move `defineProjection` imports to the SDK.

cngonzalez: ok, makes sense, thank you! something to note for our release notes (and maybe we ought to update the migration notes on the public-facing docs too)

pnpm-workspace.yaml
cngonzalez: 🤌

cngonzalez (commented):
Thank you so much for the Typegen.md rewrite, it's so clear now! 🙇  


One other question: Is there a reason for keeping `defineProjection` since it no longer does anything exported from us, really? Is it more to reduce the users' need to refactor? I worry about confusing behavior where they'll continue to expect projections to show up in their types file but it doesn't, but I really don't know what the best path / DX is here 🤔 

(on Claude calling out type collision across different `groq` packages: I'm relatively certain that those types only lived in the experimental version and aren't a danger)

ryanbonial:
Yes, `defineProjection` is mainly there to help with migration. Users can just switch the import and keep their calls and saved types. I'll mark it as deprecated in the future so that new users don't adopt it.

cngonzalez (approved):

PR #1157, opened 2026-09-02
ci(docs): add TypeDoc previews for pull requests

binoy14:
### Description

Adds pull request TypeDoc generation with a downloadable combined HTML preview artifact and a sticky generation-result comment. Keeps package-specific release uploads intact and supports optional manual uploads.

### What to review

- PR path filters for SDK source and documentation inputs
- Combined HTML artifact upload from `docs/`
- Sticky `typedoc-result` comment and artifact URL
- Release/manual package selection for JSON uploads

### Testing

- `pnpm build:docs`
- `pnpm exec oxfmt --check .github/workflows/upload-docs.yml`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/upload-docs.yml`
- GitHub “Upload Docs” workflow: TypeDoc generation, HTML artifact upload, and sticky PR comment all passed
- All 21 PR checks passed

### Fun gif

<div><a href="https://cursor.com/agents/bc-9d11d611-9ff1-4dd1-a482-cf8776ee5c5f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9d11d611-9ff1-4dd1-a482-cf8776ee5c5f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #1185, opened 2026-09-04
chore(ci): set explicit GITHUB_TOKEN permissions on workflows

binoy14:
### Description

Fixes the `actions/missing-workflow-permissions` CodeQL alerts in [code scanning](https://github.com/sanity-io/sdk/security/code-scanning) by giving each workflow an explicit `permissions` block so the `GITHUB_TOKEN` gets least privilege instead of the repo default.

- `build`, `depcheck`, `lint`, `type-check`, `test`: top-level `contents: read`.
- `test` → `report-coverage` job: adds `pull-requests: write` so the coverage action can still comment on PRs.
- `release-notes`: `permissions: {}`, it only calls a Slack webhook.

Not changed: `upload-docs` already got a permissions block in #1157, so its alert is stale. The remaining `js/incomplete-sanitization` alert is a source change and will be a separate PR.

### What to review

- Each job still has the permissions it actually uses. The shared setup action only runs `setup-node`, `pnpm/action-setup`, and `actions/cache`, which all work with `contents: read`.
- The coverage report job is the only one that writes to a PR.

### Testing

CI on this PR exercises every changed workflow except `release-notes`, which only runs on a published release. No token is used there, so an empty permissions block cannot break it.

### Fun gif

Bouncer

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


binoy14:

- #1185 ◀

ryanbonial (approved):

PR #1199, opened 2026-09-08
feat(core): stamp moduleId on message bus meta and add per-connection disconnect

joshuaellis:
### Description

An app can render more than one module at once (a main window plus a panel), and every message from it currently carries only `appId`, so the bus cannot tell the modules apart. That blocks per-module debugging, per-module reply routing, and tearing one module down without affecting its siblings.

A connection now represents one module. `connectMessageBus({appId, moduleId})` sets identity once and stamps `meta.moduleId` on every message; `moduleId` defaults to `appId`, meaning "the whole app". Each connection owns an `AbortController`, and `disconnect()` rejects its pending requests with `ABORTED`, completes its subscriptions (handler and stream form), and leaves other connections and the shared registry untouched.

### DIFF
--- a/packages/core/src/_exports/dashboard.ts
+++ b/packages/core/src/_exports/dashboard.ts
@@ -2,6 +2,7 @@ export type {
   ConnectMessageBusOptions,
   MessageBus,
   MessageBusAbortOptions,
+  MessageBusConnection,
   MessageBusEmitOptions,
   MessageBusEmitResult,
   MessageBusErrorCode,
--- a/packages/core/src/dashboard/messageBus/bus.test.ts
+++ b/packages/core/src/dashboard/messageBus/bus.test.ts
@@ -1,5 +1,6 @@
 import './__fixtures__/test-topics'
 
+import {EmptyError, firstValueFrom} from 'rxjs'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
 
 import {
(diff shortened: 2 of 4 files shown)


ryanbonial:
Items 1 and 2 from Claude seem like good things to fix before merging

joshuaellis:
I agree, I hadn't seen these before I shut my laptop for the day, thanks for flagging

joshuaellis:
> Items 1 and 2 from Claude seem like good things to fix before merging

updated.

cngonzalez (approved):
Generally looks okay to me. I investigated Claude's first issue with certain utils "dropping" a moduleId -- those utils only need to know about the app. it might be worth creating a separate type (as we've done for `MessageBus` and `MessageBusConnection` to signify what's sensitive to lifecycle and what's not) but not necessarily blocking. Let me know if I'm off and this doesn't need to be added. Otherwise I'll add a ticket for follow-up work later.

joshuaellis:
> Generally looks okay to me. I investigated Claude's first issue with certain utils "dropping" a moduleId -- those utils only need to know about the app. it might be worth creating a separate type (as we've done for `MessageBus` and `MessageBusConnection` to signify what's sensitive to lifecycle and what's not) but not necessarily blocking. Let me know if I'm off and this doesn't need to be added. Otherwise I'll add a ticket for follow-up work later.

Agree, I'll make that change in a follow up PR.

PR #1200, opened 2026-09-08
feat(core): connect one message bus per SanityInstance with module identity

joshuaellis:
### Description

A panel island and a main-window island of the same app share one module graph, so the module-level `let dashboardMessageBus` was one connection for both. Closing the panel could not disconnect only the panel, and both spoke with the same identity.

The connection is now a store keyed by `SanityInstance`: one connection per instance, carrying the federation module id the host passed down through React context (`getDashboardModuleContext`, read from the `Symbol.for('sanity.os.module')` slot the CLI wrapper populates). `DashboardTokenRefreshProvider` connects in its first render so the identity is pinned before any hook below reads the bus. The store's cleanup disconnects when the instance is disposed, so unmounting a module ends only that module's subscriptions and pending requests. When no host bus is installed the action returns `undefined` without caching.

`isDashboardEnvironment(instance)` now reads from the same store as a state source. `DashboardTokenRefreshProvider` and `AuthBoundary` pass their instance and module id; the connection is made in a `useState` initializer rather than the render body so the first `state.set` cannot notify subscribers mid-render.


packages/core/src/dashboard/messageBus/store.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Returns the message bus connection for this instance, connecting on first call.
+ *
+ * @remarks
+ * One connection is made per {@link SanityInstance}, carrying that window's
+ * module identity. The first caller wins; later callers get the cached
+ * connection. When no host bus is installed it returns `undefined` without
+ * caching, so a later call retries once a host installs the bus.
+ * @internal
+ */
+export const getDashboardMessageBus = bindActionByInstance(
```
cngonzalez, line 34: I think Claude is echoing the same thing but it might be nice to hook up the Dashboard message to the existing StateSource machinery so that you don't have to hand roll the `useSyncExternalStore` logic in the React layer; admittedly, there might be issues with this approach I'm not catching

cngonzalez (dismissed):
One note about code consistency, otherwise all makes sense to me.

I'm not familiar enough with this part of the React tree (and it's fairly new anyway) to know if Claude's comment about the lazy evaluation holds any water, but would love another person's eyes on that.

joshuaellis:
Re the review at 1d8a9d48:

1. Agreed the effect's deps could never re-trigger it, so the retry was fiction, and the test drove a transition the real store cannot make. Rather than widen the retry, removed it: the host installs the bus at module evaluation before any remote renders, and a standalone app has no host to wait for, so the connection is made once in the provider's first render. The environment state source had no remaining consumer and is gone too.

2. Intended. The CLI wrapper reconstructs the accessor itself because it must not import the SDK (SDK-2507, sanity-io/cli#1862, key fix in sanity-io/cli#1878). `module.ts` now says so and points at `render-remote.ts`.

3. The `@remarks` claim ("one connection per `SanityInstance`") is the accurate one; the PR description was loose and now says the same. No caller reads the bus from a nested provider today.

Minor: the module id is read once by design, the wrapper provides it statically above the tree; comment added. `resetMessageBus` resets the registry but does not uninstall the bus, so the tests delete the global on purpose.

joshuaellis:
Re the review at 1d8a9d48, no changes:

1. The redirect this would restore sets `window.location.href` on the host window, since federated remotes are not iframes; that takes Workbench down rather than recovering the remote. Both failure causes (no `appId`, protocol mismatch) are build or version errors, and `connectMessageBus` already warns for the first. A blank remote with a console warning is the better failure mode.

2. `ResourceProvider` holds the instance in `useState` as well (`ResourceProvider.tsx:71`), so it never changes identity under a mounted provider. The sibling's `instance` dep is for lint. Agreed the initializer is render phase either way; the comment describes why the render-phase write is safe.

3. The producer of the slot is a single known file in `sanity-io/cli`, referenced from `module.ts`. If a second tenant on the symbol shows up we'll validate it then.

4. No caller passes a `moduleId` other than the provider today. A dev warning for a mismatch can come with the first one that does.

cngonzalez (approved):
LGTM (also yes, think the Claude comment about stale instances is invalid)

PR #1212, opened 2026-09-10
feat(react): add `useTopic`

joshuaellis:
### Description

React consumers need a typed way to read dashboard message bus state topics without wiring subscriptions and Suspense themselves. `useTopic` is that read, built the same way as the SDK's other state hooks: the bus connection comes from the per-instance core store (#1200), the read state (in-flight query, its failure, subscriber refcount) lives in a per-instance core store next to it, the value is read through `createStateSourceHook`, and the hook suspends on the bus query deadline until the topic publishes.

```tsx
import {ErrorBoundary} from 'react-error-boundary'
import {useTopic} from '@sanity/sdk-react/dashboard'

function Applications() {
  const applications = useTopic('applications.list')
  return <span>{applications?.length ?? 0} applications</span>
}

  return (
    <ErrorBoundary fallback={<span>Could not load applications</span>}>
      <Suspense fallback={<span>Loading applications</span>}>
        <Applications />
      </Suspense>
    </ErrorBoundary>
  )
}
```

State topics declared with `TopicResult` resolve to their successful value. A failed result throws `TopicError` to the nearest error boundary, as does a query that passes its deadline. The recorded failure is dropped with the topic's last reader, so resetting the boundary retries with a fresh deadline, and a value published in the meantime takes precedence over it.

Event topics are not readable through this hook; `useEmit` is the event surface.

[SDK-2340](https://linear.app/sanity/issue/SDK-2340)


packages/react/src/hooks/dashboard/useTopic.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+function getMessageBus(instance: SanityInstance): MessageBus {
+  const messageBus = getDashboardMessageBus(instance)
+  if (!messageBus) throw new Error('useTopic must be used inside a dashboard application')
+  return messageBus
+}
+
+// The bus keeps one source per topic, so caching per source keeps `subscribe` stable across renders.
+const stateSources = new WeakMap<MessageBusStateSource<unknown>, StateSource<unknown>>()
```
cngonzalez: would it make sense to move the topic state and accompanying tracking into core as a global store?

cngonzalez (commented):
some concerns about logic here; is this something that makes sense to keep apart from the React lifecycle?

packages/react/src/hooks/dashboard/useTopic.ts
joshuaellis: Good idea, this is something I missed in making all this work more SDK-like. The agents guided me to be modelled similar to queryStore's shape so hopefully that's right!

joshuaellis:
Re point 2, no change:

The loop is real for a topic that publishes `undefined`, but that already violates the bus contract rather than `useTopic`'s. `MessageBusStateSource.getCurrent()` is documented as returning `undefined` before the first publish, and `query`/`firstValue` treat it the same way, so a published `undefined` is meaningless to the bus itself. Every shipped state topic uses `| null` for that reason.

Recording `resolved` on the store entry would not close it either: the suspended component never subscribes, the temporary reader is released after the grace period, and the entry drops with the flag, so the loop resumes on a 1s period. Keeping the flag would mean never dropping entries, which brings back the stale-failure problem this store exists to avoid.

cngonzalez (approved):
nice!

PR #1213, opened 2026-09-10
feat(react): add useOAuthTokens hook

joshuaellis:
### Description

Adds `useOAuthTokens`, a React hook exposing the stored OAuth token state along with `refresh` and `revoke`. It's a thin consumer of the token actions and state accessor from the core OAuth foundation (SDK-2471), continuing the move to bring SDK apps onto Sanity's OAuth authorization-code + PKCE flow (SDK-2416). `tokens` reads through core's subscription-based state source, so it stays in sync across tabs; `refresh` runs the `refresh_token` grant and `revoke` clears token state.


joshuaellis:
Actioned across the last few pushes: refresh contract test now matches core (tokens become null on the no-refresh-token path), documented the null/logout and 4xx semantics on `refresh`, added rejecting-refresh tests, wrapped the example in try/catch, dropped the unused `now` param, and removed the `useLogOut` pointer from the `revoke` docs.

Not actioning here:
- Invalid `expiresAt` -> `isExpired()` false: fixed at the root in core by #1234 (rejects unparseable `expiresAt` on deserialize), so no NaN guard in the hook.
- Moving `isOAuthTokenExpired` to core, clock-skew buffer, narrowing `tokens` to omit `refreshToken`: out of scope for this PR, will take as follow-ups if needed.

packages/react/src/hooks/auth/useOAuthTokens.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * }
+ * ```
+ *
+ * @public
+ */
+export function useOAuthTokens(): UseOAuthTokensResult {
+  const tokens = useOAuthTokensState()
+  return {
+    tokens,
+    isExpired: () => isOAuthTokenExpired(tokens),
```
ryanbonial, line 91: I think that this closure will "cache" the expired state

packages/react/src/hooks/auth/useOAuthTokens.tsx
```diff
@@ -0,0 +1,95 @@
...
+ *   )
+ * }
+ * ```
+ *
+ * @public
+ */
+export function useOAuthTokens(): UseOAuthTokensResult {
+  const tokens = useOAuthTokensState()
+  return {
+    tokens,
```
ryanbonial, line 90: I don't have a great feeling about exposing ALL of the tokens here. COuld we omit the refresh token?

packages/react/src/hooks/auth/useOAuthTokens.tsx
joshuaellis: `isOAuthTokenExpired` reads `Date.now` at call time so I'm not sure I agree – there's a test that disproves this iiuc

packages/react/src/hooks/auth/useOAuthTokens.tsx
joshuaellis: sounds good will update.

packages/react/src/hooks/auth/useOAuthTokens.tsx
ryanbonial: I (actually claude) was able to put a test together that explains what I am talking about
```ts
it('isExpired reads the current tokens, not the render-time snapshot', async () => {
  const source = createFakeTokenSource(makeTokens({expiresAt: new Date(Date.now() - 1000)}))
  mockGetState.mockReturnValue(source)
  const fresh = makeTokens({accessToken: 'new'})
  mockRefresh.mockImplementation(() => {
    source.set(fresh)
    return Promise.resolve(fresh)
  })
  const {result} = renderHook(() => useOAuthTokens(), {wrapper})
  const {isExpired, refresh} = result.current
  expect(isExpired()).toBe(true)
  await act(async () => {
    await refresh()
  })
  expect(result.current.isExpired()).toBe(false) // fresh render: passes
  expect(isExpired()).toBe(false) // captured closure: fails, still true
})
```

joshuaellis: Thanks! Will look at fixing 🥸 TDD ftw

joshuaellis: fixed!

ryanbonial (approved):

PR #1216, opened 2026-09-11
feat(react): add useHandleOAuthCallback hook

joshuaellis:
### Description

Adds `useOAuthCallback`, the OAuth counterpart to `useHandleAuthCallback`


joshuaellis:
Both actioned: renamed to `useHandleOAuthCallback` to match `useHandleAuthCallback`, and dropped the `AuthBoundary` sentences from the TSDoc; they return in #1223 where `handleAuthCallback` starts delegating to the OAuth path.

ryanbonial (approved):

PR #1217, opened 2026-09-11
feat(react): add useOAuthAuthorize hook

joshuaellis:
### Description

Adds `useOAuthAuthorize`, a hook that starts the OAuth authorization-code + PKCE flow by delegating to core's `startOAuthAuthorization`. It returns the bound action directly, matching `useOAuthCallback` and `useLogOut`, so apps get one consistent shape for auth callbacks. Config (`clientId`, `redirectUri`, `organizationId`) is read from the instance's `auth.oauth`, so the hook takes no options.


joshuaellis:
Actioned: reworded to "the returned promise rejects" and the example now does `authorize().catch(console.error)`. Also updated the `{@link}` to `useHandleOAuthCallback` following the rename in #1216.

binoy14 (approved):

ryanbonial (approved):

PR #1223, opened 2026-09-11
feat(react): drive OAuth flow from AuthBoundary

joshuaellis:
### Description

A developer opts into Sanity OAuth by adding `auth.oauth` to their `SanityApp` config and `AuthBoundary` handles the rest: logged-out users are sent to the authorize endpoint, the callback is completed and the URL cleaned, children render once authenticated, and logging out revokes the tokens. Apps without `auth.oauth` are unaffected.


joshuaellis:
Actioned:
- (1) `getOauthInitialState` now only treats `code`/`error` as a callback, so a stray `?state=` boots to `LOGGED_OUT` instead of hanging in `LOGGING_IN`. Test added.
- (2) authorize rejections are stored as `{error}` so a falsy reason still surfaces; the existing rejection test now rejects with `undefined`.

Not actioning:
- (3) The effect already excludes studio/dashboard before checking `isOAuth`, so the two agree today; will revisit if `resolveAuthMode` precedence changes.
- (4) Logout bouncing to authorize matches the legacy `loginUrl` behaviour and is intended for now. Test-order coupling: not causing failures, leaving as is.

ryanbonial (approved):

PR #1234, opened 2026-09-11
fix(core): reject persisted OAuth tokens with an unparseable expiresAt

joshuaellis:
### Description

A corrupt `expiresAt` in stored OAuth tokens no longer reads as a token that never expires. Such an entry is treated as no stored token and removed, so the user goes through the login flow again instead of sitting on a token nothing will ever refresh.


ryanbonial:
@joshuaellis how would an invalid `expiresAt` get put into this token object? `expiresAt` is something that we create [here](https://github.com/sanity-io/sdk/blob/feat/SDK-2538/packages/core/src/auth/oauth/oauthActions.ts#L79) with the actual `expires_in` response. So, it seems like we should verify that `expires_in` is valid when we get it and reject it then instead of so late in the token process. If a token has a valid `expiresIn` but an invalid `expiresAt` something wonky happened. 

I should mention that I'm not against adding this defensive check here, just that adding another higher up seems more important.

joshuaellis:
Nothing to action here. (1) The cross-tab path self-heals on next load, leaving as is. (2) Forcing re-login on a corrupt entry is intended; not worth a partial-recovery path for a case that shouldn't occur in practice.

joshuaellis:
It can't come from `toOAuthTokens` in practice: a bad `expires_in` makes `serializeTokens` throw on `toISOString()` (`RangeError: Invalid time value`), so it never reaches storage. The case this covers is storage written by something other than us (older SDK shape, other scripts on the origin, manual edits), and `deserializeTokens` is the one place all reads go through. Happy to add a guard on `expires_in` for a clearer error than the `RangeError`, but as a separate change.

beep boop im a bot

binoy14:
Don't have context, ping me if my review is still needed

packages/core/src/auth/oauth/oauthAuth.ts
```diff
@@ -68,7 +72,12 @@ export function getOauthInitialState(options: AuthStrategyOptions): AuthStrategy
   const redirectUri = authConfig.oauth?.redirectUri
 
   // Persisted tokens win
-  const tokens = deserializeTokens(storageArea?.getItem(OAUTH_TOKENS_KEY) ?? null)
+  const rawTokens = storageArea?.getItem(OAUTH_TOKENS_KEY) ?? null
+  const tokens = deserializeTokens(rawTokens)
+  if (rawTokens && !tokens) {
```
ryanbonial, line 77: One similar thing I noticed here is that if the `expiresAt` is already in the past, it would start as `LOGGED_IN` but likely not use the refresh token because it would hit an endpoint and then go through the whole oauth back and forth.

joshuaellis: Yes good point, solved this by changing the flows:

```
Expired + refresh token → starts LOGGING_IN, refreshes on init. 
Expired without → dropped like corrupt.
```

ryanbonial (approved):

PR #1235, opened 2026-09-11
fix(core): ignore stale OAuth callbacks once a session is established

joshuaellis:
### Description

Re-running the OAuth callback with a URL whose params have already been exchanged (for example a custom callback component remounting before the URL was cleaned) no longer throws the user out of an established session into an error state. Validation failures while logged out still surface as errors.


packages/core/src/auth/oauth/oauthActions.ts
ryanbonial: Do we want to add a call to `rejectCallback` in this case were an error was returned from the server, but the user is currently logged in?

joshuaellis:
Actioned (1) and (2): `clearOAuthArtifacts` now only runs on the rejecting branch, with the existing stale-callback test extended to assert re-auth artifacts survive; TSDoc updated to say a string return does not imply an exchange ran. (3) is out of scope here, agree it's a follow-up.

packages/core/src/auth/oauth/oauthActions.ts
joshuaellis: Good call, same rule should apply. Will route the `error` branch through `rejectCallback` too.

ryanbonial (dismissed):

ryanbonial (approved):

PR #1246, opened 2026-09-14
refactor(core): give installMessageBus its own options type without moduleId

joshuaellis:
### Description

`installMessageBus` reused `ConnectMessageBusOptions`, which carries `moduleId` since #1199, but the host has only an app identity and the field was silently dropped. It now takes its own `InstallMessageBusOptions` with just `appId`, so a caller cannot pass a value that does nothing. This mirrors the `MessageBus` / `MessageBusConnection` split between the shared bus and a lifecycle-bound connection.


cngonzalez (dismissed):

joshuaellis:
sorry @cngonzalez had to rebase.

cngonzalez (approved):

PR #1247, opened 2026-09-14
feat(sdk-react): expose federation setup

gu-stav:
### Description

The module-federation setup currently lives in `@sanity/workbench`, so SDK consumers cannot create dashboard remote clients without depending on Workbench.

Expose `createRemoteInstance` through `@sanity/sdk-react/dashboard` with the runtime surface needed to register, load and preload remotes. The caller supplies the instance name, which also names its SDK logger so lifecycle logs remain attributable when several remote hosts are active.

### What to review

- Does the minimal runtime surface cover the dashboard use cases?
- Should module ids use a branded type rather than `string`?

### Documentation

Added TSDoc for `createRemoteInstance` and its public types.


packages/react/src/dashboard/createRemoteInstance.ts
```diff
@@ -0,0 +1,124 @@
...
+  /** Preloads assets without evaluating them; omit exposes to warm the whole remote. */
+  preloadRemote(name: string, exposes?: string[]): Promise<void>
+}
+
+/**
+ * Creates a client for registering, loading, and preloading federated modules.
+ * @public
+ */
+export function createRemoteInstance(options: CreateRemoteInstanceOptions): RemoteInstance {
+  const logger = createLogger('sanity-remote')
```
joshuaellis, line 50: it would be be better if we pass a name to the create so the logger is differable between remotes i think?

gu-stav: Good idea 👍🏼 Updated - `createRemoteInstance` now requires a name and both the instance and the logger are using it.

joshuaellis (dismissed):

joshuaellis (approved):

PR #1248, opened 2026-09-14
feat(sdk-react): add lazy instance-scoped `useRemoteClient`

gu-stav:
### Description

Dashboard React consumers need one module-federation client per Sanity instance, created only when a hook first requests it.

Expose `useRemoteClient` through `@sanity/sdk-react/dashboard`. It uses the same `createStateSourceHook` path as the other SDK hooks and keeps each instance's state source in a `WeakMap`, so core store primitives do not need to become part of the `_internal` contract.

### What to review

- Does the state-source lifecycle match the other lazy SDK hooks?

### Documentation

Added TSDoc for `useRemoteClient`.


gu-stav:
> The federation host is never torn down (`remoteClientStore.ts:9-14`)

There is no way to destroy a module-federation instance.

packages/react/src/hooks/dashboard/useRemoteClient.ts
```diff
@@ -0,0 +1,12 @@
+import {type createRemoteInstance} from '../../dashboard/createRemoteInstance'
+import {getRemoteClient} from '../../dashboard/remoteClientStore'
+import {useSanityInstance} from '../context/useSanityInstance'
+
+/**
+ * Returns a remote client shared by hooks using the same Sanity instance.
+ * The first call creates its store; later calls reuse the client.
+ * @public
+ */
+export function useRemoteClient(): ReturnType<typeof createRemoteInstance> {
```
joshuaellis, line 10: I think this should be using `createStateSourceHook`, was there a reason not too?

joshuaellis (changes requested):

packages/react/src/hooks/dashboard/useRemoteClient.ts
gu-stav: Updated 👍🏼

joshuaellis (approved):

PR #1250, opened 2026-09-14
feat(core): export internal mutation evaluator

ryanbonial:
### Description

Export the core mutation evaluator for server-side use without requiring a client, document store, or listeners. This keeps server code and the SDK on the same mutation implementation instead of duplicating mutation semantics.

The export includes `processMutations`, its options and document types, and the underlying patch operations. These APIs remain marked `@internal` because persistence, authorization, and transaction handling stay with the caller.

[SDK-2549](https://linear.app/sanity/issue/SDK-2549/export-core-mutation-evaluator-for-server-side-use)

### What to review

Review the new `@sanity/sdk-core` export surface and the API documentation in `processMutations.ts`. Mutation behavior itself is unchanged.

### Testing

No tests were added because this exposes the existing evaluator implementation without changing its behavior. The SDK continues to use the same implementation and existing tests.

### Documentation

No public documentation is required. The exported APIs are marked `@internal`.

### Fun gif
psst


packages/core/src/document/processMutations.ts
```diff
@@ -108,18 +131,29 @@ export function getDocumentIds(selection: MutationSelection): string[] {
...
- * Applies the given mutation to the given document set. Note, it is expected
- * that all relevant documents that the mutations affect should be within the
- * given `document` set. If a document does not exist, it should have the value
- * `null`. If a document is deleted as a result of the mutations, it will still
- * have its document ID present in the returns documents, but it will have a
- * value of `null`.
+ * Applies mutations to an in-memory document set.
+ * Deleted documents remain in the returned set with a value of `null`.
+ *
+ * Returns the input map when `mutations` is empty. Results can share document
```
cngonzalez: What is "the input map" in this case? The `DocumentSet` type of `{[documentId]: Document}`?

packages/core/src/document/processMutations.ts
```diff
@@ -108,18 +131,29 @@ export function getDocumentIds(selection: MutationSelection): string[] {
...
+ * and nested-value references with the input.
```
cngonzalez: What does this mean -- that the DocumentSet also includes any reference extensions included in the input? Or that there's some object equality thing going on?

packages/core/src/document/processMutations.ts
```diff
@@ -108,18 +131,29 @@ export function getDocumentIds(selection: MutationSelection): string[] {
...
+ *
+ * Sets `_rev` to `transactionId` on created, replaced, and patched documents.
+ * Uses `timestamp` for timestamps unless the mutation preserves them.
+ * Ignores `merge` patches; `inc` and `dec` skip values that are not numbers.
+ * Generates document IDs when omitted or ending in `.`.
+ * Patch operations can generate missing array keys.
+ *
+ * @remarks
+ * Uses the SDK's evaluator so callers share its mutation semantics without
+ * starting a client or document store. Does not run Content Lake's server
```
cngonzalez: Would something like "runs SDK mutations over an in-memory set of document" be a little clearer? Feel like this comment is geared a little too much toward the specific use case we just received.

cngonzalez (changes requested):
some clarifying questions / comments on the comment. I agree with Claude that threading through the generic would be nice; not sure if it's appropriate in this PR?

ryanbonial:
Updated comments, ready for another review. Thanks!

cngonzalez (approved):
Thank you! 🙏

PR #1251, opened 2026-09-14
feat(core): give each message bus connection its own state

joshuaellis:
### Description

A state topic has one shared value that every connected application reads. That works for `navigation.location` and breaks per-app `auth.token`: one shared value cannot hold a different token per app.

Each connection now owns its own copy of every state topic. `subscribe`, `query`, `getCurrent` and `useTopic` read the caller's copy only. A copy starts at the topic's manifest seed and stays unpublished until the host writes to it.

The host writes to one connection at a time through a `MessageBusClient` from `host.connections`. The observable emits every open application connection, then each new one as it joins. Each connection gets one client object with a `closed` signal. There is no broadcast call. Same value for everyone is a loop over clients; current values for a new connection are written when it appears on `connections`.

```ts
const os = installMessageBus({appId: 'dashboard'})

os.connections.subscribe((client) => {
  client.emit('auth.token', tokenFor(client.appId))
  client.emit('navigation.location', currentLocation)
})
```

`useTopic`, `useEmit`, `TokenRefreshProvider` and `connectMessageBus()` are unchanged.


joshuaellis:
**Beep Boop, I'm a bot.**

1. Agreed. `connections` now excludes only the host's own record by identity (4e1a2e5); a sibling connection sharing the host app id is listed and writable. Test added.
2. Not treating the message bus as a compatibility surface yet; the title stays `feat`.
3. That is the model from SDK-2502: a connection's lifetime is its `SanityInstance`, and `getDashboardMessageBus` disconnects on dispose. Added to the `connections` TSDoc.
4. Nothing published `preferences.*` or `panels.mode` from an application; every state emit in workbench is the host's. They are host-resolved reads.

`installMessageBus` dropping `moduleId` is handled in #1246. `fallow audit`, `ts:check` and tests pass locally.

gu-stav (approved):

PR #1252, opened 2026-09-15
fix(release): bump rc versions without pnpm version

joshuaellis:
The release-candidate job fails on the second package bump with `ERR_PNPM_UNCLEAN_WORKING_TREE`.

pnpm 12 (adopted in #1208) turned `pnpm version` into a native workspace command that syncs the lockfile and refuses to run on a dirty git tree — the script's assumption that it is "just an alias for npm version" no longer holds. Because `release-rc.mts` bumps packages one at a time, the first bump dirties the tree and the second aborts.

Writing the `version` field directly to each `package.json` sidesteps the native command's install and clean-tree checks. Cross-package refs use `workspace:*`, so no dependent ranges or lockfile entries need updating.


cngonzalez (approved):
no spinner? we hate fun?

PR #1257, opened 2026-09-16
ci(release): replace release-please with changesets

joshuaellis:
### Description

Releases are driven by Changesets instead of release-please. A PR that should ship adds a `.changeset/*.md` (`pnpm changeset`; feat = minor, fix/perf/revert = patch, breaking = major). No changeset, no release.

On every push to `main`, `release.yml` runs `changesets/action`, which opens or updates a `chore: release` Version PR with the ECOSPARK app token so CI runs on it. Merging the Version PR consumes the changesets and bumps both packages. The next run on `main` finds no changesets, builds, runs `pnpm test:pack`, and publishes `@sanity/sdk` and `@sanity/sdk-react` to `latest` with provenance. Both packages share a version (`fixed` group). Private workspace packages are not versioned.

The workflow creates one `sdk-vX.Y.Z` tag and one GitHub release per version. The release body is the new CHANGELOG section of each package. `release-notes.yml`, the Linear "Shipped" step and the CircleCI trigger key off that release as before. `upload-docs.yml` parsed the component out of a per-package tag; it now reads the version from the single tag and uploads docs for both packages.

Publishing is idempotent. If both packages at the current version are already on the registry the job skips publish and exits green. `changeset publish` skips packages already on the registry, so a rerun after a partial publish republishes only the missing one. The job fails if the action's `has-changesets` output is missing, so a renamed output cannot leave the publish path silently skipped.

The topmost heading in each CHANGELOG changed from `## [3.3.0](compare-link)` to `## 3.3.0 (date)`. Changesets inserts new entries before the first heading matching `^## \d+\.\d+`; the bracketed form does not match, so the first entry would land above `## 0.0.0` at the bottom. Older headings are unchanged.

`scripts/release-rc.mts` and `scripts/release-branch.mts` read the deleted release-please config and are replaced in the next PRs in this stack. Do not dispatch `release-rc.yml` until the rc PR lands. Release-please already shipped 3.3.0 (#1254), so there is no backfill changeset; the next release comes from changesets added after this lands.


joshuaellis:
**beep boop, I'm a bot**

Actioned from the automated review:

- 3: the registry guard now requires both packages to be present before skipping publish.
- 4: added a step that fails if the action's `has-changesets` output is empty. The kebab-case names were verified against the `action.yml` at the pinned SHA.

Not actioned:

- 1: `release-rc.mts` / `release-branch.mts` are replaced in #1258 and #1259. Stated in the description; do not dispatch `release-rc.yml` until #1258 lands.
- 2: Renovate changesets are #1261.
- 5: `RENOVATE_STRATEGY.md` and the dashboard header are rewritten in #1261.
- `$schema`: `@changesets/config@4.0.1` is what `@changesets/cli@3.0.3` installs.

Also: `origin/main` gained #1256 (`feat`) since this was opened, so the backfill changeset is now `minor` and lists it.

binoy14 (dismissed):

binoy14 (approved):

PR #1261, opened 2026-09-16
ci(renovate): write a patch changeset for production dependency bumps

joshuaellis:
### Description

Renovate PRs that bump a production dependency of `packages/core` or `packages/react` carry a patch changeset and release through the Version PR flow. Dev, tooling and app bumps carry none.

Two new package rules add a `postUpgradeTasks` hook running `pnpm changeset:renovate <branch> <dep@version>...`: one scoped to the two published `package.json` files for inline deps, and one matching the catalog runtime deps of core/react by name (`@sanity/client`, `@sanity/comlink`, `@sanity/message-protocol`, `@sanity/types`, `groq`, `groq-js`, `rxjs`, `react-compiler-runtime`, `react-error-boundary`). Catalog deps report `depType: pnpm.catalog.default` and `packageFile: pnpm-workspace.yaml`, so a file-scoped rule alone misses them. The script (`scripts/changeset-renovate.mjs`, Node built-ins only, plain JS so it does not depend on the Renovate container's Node version) writes `.changeset/renovate-<slug>.md` with a patch bump for both packages and a body listing the updated deps. The filename comes from the branch, so a grouped PR gets one file and a rebase overwrites it. `executionMode: branch` runs the hook once per branch with all upgrades in it. The hook is not on the `@sanity/*`, `dependencies` and `groq` grouping rules, so an apps-only or dev-only bump of a shared dep does not trigger an SDK release.

Commit type no longer drives releases; the production rules move from `fix(deps)` to `chore(deps)`. `renovate.yml` allows the new command. `RENOVATE_STRATEGY.md`, the dependency dashboard header and `.claude/commands/review.md` describe changesets instead of `fix(deps)`. A new catalog runtime dep added to core/react must also be added to the name list or its bumps will not release; the `/review` command flags this.

Not verified locally: whether Renovate applies the hook in a grouped branch where only some upgrades match the scoped rule. Check on the first production-dep PR after merge.


joshuaellis:
**beep boop, I'm a bot**

Actioned from the automated review:

- 1 and 2: added a second rule matching the catalog runtime deps of core/react by name (`@sanity/client`, `@sanity/comlink`, `@sanity/message-protocol`, `@sanity/types`, `groq`, `groq-js`, `rxjs`, `react-compiler-runtime`, `react-error-boundary`). The file-scoped rule stays for inline deps. `RENOVATE_STRATEGY.md` and `review.md` updated; the caveat is now "a new catalog runtime dep must be added to the name list".
- 3: script renamed to `.mjs`, run with plain `node`.

Not actioned:

- 4: `installTools` and `{{#each upgrades}}` in branch mode cannot be verified locally; noted in the description, confirm on the first production-dep PR after merge.

binoy14 (approved):

PR #1264, opened 2026-09-16
feat(react): make useOrganizationId and useWindowTitle work under the message bus

joshuaellis:
### Description

Makes `useOrganizationId` and `useWindowTitle` from `@sanity/sdk-react/dashboard` work in both Dashboard runtimes. Each hook selects its transport internally: the message bus when a host has installed one (`isDashboardEnvironment()`), otherwise the existing Comlink path. Public signatures and iframe behaviour are unchanged.

`useOrganizationId` reads the `organizations.current` topic under the bus. `useWindowTitle` stays Comlink-only but no-ops under the bus instead of suspending forever on `useWindowConnection`, since the host owns the document title there.

The transport branch in `useWindowTitle` is a deliberate rules-of-hooks exception: the Comlink hook suspends and must not run under the bus, and the branch is constant for the page lifetime. `eslint-plugin-react-compiler` reports any `react-hooks/*` disable as its own non-suppressible error, hence the file-level disable of that rule.


packages/react/src/hooks/dashboard/useWindowTitle.ts
```diff
@@ -66,6 +71,15 @@ function resolveAppTitle(resource: ContextResource): string | undefined {
  * @public
  */
 export function useWindowTitle(viewTitle?: string): void {
+  // The document title is set by the Dashboard host under the message bus, so this hook only
+  // manages it in the Comlink runtime. The branch is stable: the transport is fixed for the page
+  // lifetime, so the Comlink hooks below always run or never run.
+  if (isDashboardEnvironment()) return
```
gu-stav, line 77: Should we `console.warn` in this case?

gu-stav (approved):

packages/react/src/hooks/dashboard/useWindowTitle.ts
joshuaellis: i think its probably fine not too, we can revisit if we get feedback on this?

gu-stav: No strong opinion, but with a warning users have a feedback at least about this being a no-op. Now that I think about it more - does changing a document in studio for example update the page title? Dashboard only sets the initial title I think which makes me wonder whether this hook should set the title even under the message?

joshuaellis: > does changing a document in studio for example update the page title?

idk

PR #1266, opened 2026-09-16
feat(dashboard): add new topic `applications.status.update`

gu-stav:
### Description

Register a new event topic `applications.status.update` that is hidden from the public TS interface of the message bus. We will use this topic to set the unread badge for an application in the dock. It stays invisible until we have validated the idea.

It was originally implemented in https://github.com/sanity-io/workbench/pull/643 but wasn't part of the port.

### Documentation

No public documentation because the topic is internal.


gu-stav:
@claude review

joshuaellis (dismissed):

joshuaellis (approved):

PR #1270, opened 2026-09-17
feat(react): make useStudioWorkspacesByProjectIdDataset work under the message bus

joshuaellis:
## Description

`useStudioWorkspacesByProjectIdDataset` is Comlink-only today: it fetches `dashboard/v1/context` over `useWindowConnection`, which never resolves in the federated Dashboard runtime, so the hook (and `useNavigateToStudioDocument` on top of it) suspends forever there.

The hook now selects its transport with `isDashboardEnvironment()`, the same way `useWindowTitle` does. Under the message bus it derives the workspace map from the `applications.list` topic: each workspace of a studio's active deployment becomes one resource, mirroring the shape the legacy Comlink protocol returns. The Comlink path and the public signature are unchanged.

`applicationOrigin` moves from `useApplications` into core as `getApplicationOrigin` so both hooks resolve a studio's URL the same way.


packages/react/src/hooks/dashboard/useStudioWorkspacesByProjectIdDataset.ts
```diff
@@ -24,8 +28,13 @@ interface StudioWorkspacesResult {
   error: string | null
 }
 
+type DashboardApplications = NonNullable<TopicData<'applications.list'>>
```
gu-stav, line 31: Why make it non-nullable?

gu-stav (approved):
Small clarifying question but otherwise looks good.

packages/react/src/hooks/dashboard/useStudioWorkspacesByProjectIdDataset.ts
joshuaellis: Because its only used as a type for the helper functions to avoid having to handle nullish in there.

PR #1274, opened 2026-09-17
feat(react): make useNavigate and useNavigateToStudioDocument work under the message bus

joshuaellis:
## Description

`useNavigate` and `useNavigateToStudioDocument` sit on `useWindowConnection`, so they suspend forever in the federated Dashboard runtime. Both now pick their transport with `isDashboardEnvironment()`, like `useWindowTitle`; the Comlink paths are unchanged.

`useNavigate` also returns a function for reporting the app's own router location to the Dashboard. Federated apps need this because the app and the Dashboard are two routers on one window. Under Comlink it sends the `update-url` bridge message.

`MessageBusConnection` exposes `appId` so an app can tell its own navigations apart on the bus.


joshuaellis:
**beep boop, I'm a bot**

Addressed 1, 3 and 4b in 49f559ce. Two findings I'm not acting on:

**2. Host navigations echoed back.** The redundant `navigation.location.update` is real, but the duplicate history entry isn't. Workbench's navigation machine treats a request for the current href as a no-op (`machine.ts:189`), and the Comlink host applies `update-url` as `replace`. So the echo is a message the host discards in both runtimes. Suppressing it would add a second ref and a second comparison to protect against nothing.

**4a. Mount-time `push`.** Same reason. The host's same-href guard means `type` on the mount sync has no effect on the bus, and Comlink ignores `type` entirely. `push` and `replace` are indistinguishable there.

gu-stav (approved):

PR #1275, opened 2026-09-17
feat(dashboard): let the host refuse bus events and state topics per connection

joshuaellis:
### Description

The Workbench host is about to ignore bus traffic from applications that are not on screen (SDK-2467): a backgrounded window or a closed panel must not act on the host, and must not be written state while hidden. The bus had no way to say so. A responder could only `reply()` or throw (which surfaces as `HANDLER_THREW`), and a host could only write a state topic to a connection, never decline it, so a hidden application's `query()` would hang until `TIMEOUT`.

This adds an explicit rejection lane:

- `REFUSED` joins `MessageBusErrorCode`.
- `MessageBusMessage.reject(reason?)` settles a pending event reply so the sender's awaited `emit()` rejects with `MessageBusError('REFUSED', reason)`.
- `MessageBusClient.reject(topic, reason?)` rejects a state topic for one connection: pending and later `query(topic)` reject with `REFUSED` until the host next writes that topic to the connection. Subscriptions never error, so they resume on the next write.

The rejection is tracked beside the topic's subject rather than by erroring it, because a rejected promise cannot be un-rejected and an errored subject cannot resume.


gu-stav:
@joshuaellis If applications e.g. in the background don't receive all messages: what for example would happen if a user changes the color theme and the applications come into the foreground again? Would it then replay the missed events? Or does it read the latest state from all state topics when it comes into foreground again?

joshuaellis:
> @joshuaellis If applications e.g. in the background don't receive all messages: what for example would happen if a user changes the color theme and the applications come into the foreground again? Would it then replay the missed events?

yeah itll need to be replaying events. This ticket SDK-2468, touches upon it, i widened 2467 to be more "ignore everything" and will most likely do the same on 2468 when I get to it :)

packages/core/src/dashboard/messageBus/bus.ts
```diff
@@ -207,6 +214,19 @@ export interface MessageBusClient {
...
+   * writes a value with {@link MessageBusClient.emit}, which clears the refusal.
+   *
+   * @remarks
+   * A rejected promise cannot be un-rejected and an errored subject cannot resume, so the refusal
+   * is tracked per connection alongside the topic's subject rather than by erroring it: `query`
+   * consults the refusal before awaiting, and {@link MessageBus.subscribe} and the topic's
+   * {@link MessageBusStateSource} stay silent (they never error) until a value is written.
+   * Ignored once the connection has closed, the same as {@link MessageBusClient.emit}.
+   */
+  refuse<K extends StateTopic>(type: K, message?: string): void
```
gu-stav: Should this be called `reject` for naming consistency? Leaving refuse is fine for me, but then we should probably rename it in `bus.ts`?

joshuaellis: can do. I will make the change.

joshuaellis: updated.

gu-stav:
@joshuaellis Looking at Claude's review comment: 1 seems relevant: 

```
1. `reset()` leaks rejections across a bus reset — bus.ts:1172

reset() clears every other piece of per-connection state but not rejections:

for (const connection of registry.connections) {
  for (const subject of connection.stateSubjects.values()) subject.complete()
  connection.stateSubjects.clear()
  // rejections survives
}

After a reset, resolveStateSubject rebuilds the topic's subject from its manifest seed, but query() still short-circuits on connection.rejections.has(type) and rejects with REFUSED for a topic that now has a perfectly good seed value. The only escape is a client.emit() for that exact topic. resetMessageBus is @internal test isolation, so the blast radius is consumer test suites leaking REFUSED between tests — still wrong, and cheap to fix.

Add connection.rejections.clear() alongside connection.stateSubjects.clear().
```

joshuaellis:
**beep boop, I'm a bot**

Good catch, that was a real leak: `reset()` dropped every state subject but left `rejections` in place, so a `REFUSED` survived into the next test. Fixed in 13ee97d5: `reset()` now clears `rejections` alongside `stateSubjects`, with a test that rejects a topic, resets, and checks `query()` resolves to the manifest seed again.

gu-stav (approved):

PR #1280, opened 2026-09-18
fix(core): scope a failed document read to the document it belongs to

gu-stav:
### Description

One `documentStore` is shared by every document in a project and dataset. A failed read could either break every document in that store or leave a non-`liveEdit` read loading forever when its draft is inaccessible.

The shared store error caused the first failure:

```ts
export interface DocumentStoreState {
  documentStates: {[TDocumentId in string]?: DocumentState}
  error?: unknown // one error for the whole resource
}

// every per-document listen() stream is merged into this one subscription
.subscribe({error: (error) => state.set('setError', {error})})

selector: ({state: {error, documentStates}}, options) => {
  const documentId = DocumentId(docId)
  if (error) throw error // thrown for every document in the resource
```

The draft and published lookup caused the second:

```ts
const draft = documentStates[getDraftId(documentId)]?.local
const published = documentStates[getPublishedId(documentId)]?.local
// an error under `drafts.<id>` is never read, so this returns forever
if (draft === undefined || published === undefined) return undefined
```

Both happen when `/doc/` omits a document for a reason other than non-existence, including insufficient permissions.

The draft failure affected the default `useDocument` path. A user who could read the published document but not its draft saw the document remain in its loading state with no error surfaced. At the same time, `getDocumentSyncStatus` reported it as synced.

Errors are now scoped to the document that failed. Readable documents continue to resolve, while an inaccessible draft rejects instead of loading forever.


gu-stav:
@claude mark as ready for review (sorry can't do from the GH app).

gu-stav:
@claude review again

gu-stav:
> 3. Medium — getPermissionsState still loads forever

> // packages/core/src/document/permissions.ts:93-98
> const local = documentStates[documentId]?.local
> if (local === undefined) return undefined

> No error check, and calculatePermissions doesn't read state.error either. useDocumentPermissions for a document with an inaccessible draft stays pending forever — same bug this PR fixes for useDocument. Worth reusing the new per-document error check here.

#3 (Medium) — real but pre-existing, not touched. getPermissionsState returns undefined while local === undefined and never consults the error — before my change too, since it never read state.error either. It hung on main in exactly the same way. Fixing it is the same one-line shape, but it's not a regression and it's outside the minimum.

I'll create a follow-up issue.

cngonzalez (approved):
logic LGTM -- I'd recommend the Claude recommendation on importing `getDocumentIdFromDocumentHandleLikes` -- the retry-on-new-subscriber also sounds nice, but I'd accept that as a possible future improvement

gu-stav:
@cngonzalez Good idea! Updated in c8c1d82828dda5e68f264080c1b96ec733921bec ✅

cngonzalez (approved):

PR #1281, opened 2026-09-18
fix(core): read a document error from every id the selector reads

gu-stav:
### Description

A read that is not `liveEdit` resolves a draft and a published document, and returns early while either of them is missing. The error was only ever read from the requested id:

```ts
selector: ({state: {error, documentStates}}, options) => {
  const documentId = DocumentId(docId)
  if (error) throw error
  throwDocumentError(documentStates, documentId) // only the requested id

  if (liveEdit) {
    document = documentStates[documentId]?.local
  } else {
    const draft = documentStates[getDraftId(documentId)]?.local
    const published = documentStates[getPublishedId(documentId)]?.local
    // an error under `drafts.<id>` is never read, so this returns forever
    if (draft === undefined || published === undefined) return undefined
```

So an inaccessible draft was recorded under `drafts.<id>`, where neither selector looked. `useDocument` then suspended indefinitely with nothing thrown and nothing logged, and `getDocumentSyncStatus` was worse: the draft's state entry exists (it holds the error), so its own early return never fired and it reported the document as synced.

A reader who may read published documents but not drafts is an ordinary Sanity configuration, and it is exactly the case where the published read succeeds and the draft read fails. Each selector now reads the error from every id it resolves — version under a release perspective, draft and published, or the single id under `liveEdit`.

Follows #1280, which recorded the error per document but left this half of it. SDK-2602 documents both.

### What to review

`throwDocumentError` is variadic now, and each selector calls it next to the ids it is about to read rather than once up front. That keeps the guard beside the read it protects, and keeps the selectors at the complexity they had before #1280.

### Testing

`documentStore.test.ts` gains `reports an inaccessible draft to the document that reads it`, which asserts both selectors through their public surface:

```ts
await expect(resolveDocument(instance, doc)).rejects.toThrow(message)
await expect(
  firstValueFrom(
    getDocumentSyncStatus(instance, doc).observable.pipe(first((value) => value !== undefined)),
  ),
).rejects.toThrow(message)
```

`resolveDocument` is what `useDocument` suspends on, so the first assertion exercises the path consumers actually take. Without this change the first assertion times out, which is the defect: the document never settles. Removing only the `getDocumentSyncStatus` guard fails the second in 20ms with `promise resolved "true" instead of rejecting`.

The expected message carries the draft id in backticks, so a fix that threw the published document's error instead would not match.

Two cases stay untested and are cheap to add later if wanted: the symmetric inaccessible-published read, and the release-perspective id. Both go through the same helper as the covered cases.

### Documentation

No public API change.


gu-stav:
Folded into #1280 — both halves of SDK-2602 ship together there.

gu-stav:
Folded into #1280. Recording the error per document only pays off if the selector reads it from the right id, so the two halves are one change and read better reviewed together. #1280 now covers both cases and carries the test from here.

